PR #4, opened 2022-12-19
remove ini config parsing

JakeHillion:
## Summary

Remove the `.ini` config parsing. This has been deprecated for a while and the old style is pretty much useless with the new type stuff. Removing now so no open source users start using it.

## Test plan

- `make test-devel`
- CI?


jgkamat (approved):
Looks good to me, thanks for the fix!

PR #5, opened 2022-12-19
Add autodeployment for website

jgkamat:
## Summary
Add autodeployment for website

The last commit is a hack to fix things up on my fork for testing, which dosen't need to go into the main repo.

Yarn was also complaining to me about a package lock file, so I complied with those complaints by switching to a yarn lock.

This also deploys to `gh-pages` primarily because I can't figure out how to change it, and secondarily because it's easy to revert if it messes up!

## Test plan
https://jgkamat-website-test.github.io/object-introspection/

https://github.com/jgkamat-website-test/object-introspection/actions/runs/3735828967/jobs/6339506038


website/docusaurus.config.js
```diff
@@ -14,7 +14,7 @@ const config = {
   title: 'Object Introspection',
   tagline: 'Dynamic C++ Object Profiling',
   url: 'https://your-docusaurus-test-site.com',
-  baseUrl: '/',
+  baseUrl: '/object-introspection/',
```
tyroguru, line 17: I pushed a fix for this yesterday so you can restore this file before landing.

.github/workflows/deploy.yml
```diff
@@ -0,0 +1,44 @@
+name: Deploy to GitHub Pages
+
+on:
+  push:
+    branches:
+      - jay/website
```
tyroguru, line 6: I assume this should just reference the main branches "website" directory and not yours?

tyroguru (approved):
Jay I've approved it with 2 comments. The push branch is the main one for me.

.github/workflows/deploy.yml
jgkamat: These two changes you commented on shouldn't be committed, they're in the top "revert me" commit. They are changes required to test the change on my fork but would break things on the actual build.

PR #7, opened 2022-12-20
remove funcgeninternal.h

JakeHillion:
## Summary

The container information added by `FuncGenInternal.h` has been made redundant by the pluggable containers feature. Remove this legacy difference.

## Test plan

- `make test-devel`
- CI
- D42167455

PR #8, opened 2022-12-20
container regex matching

JakeHillion:
## Summary

Add the option to provide a custom regex in a container info file. This allows us to remove the hyper specific if conditions in the matching code, instead specifying them with their containers. In future it should also allow us to match multiple paths in the same container info file, reducing duplication.

## Test plan

- `make test-devel`
- CI

This needs to go in after #7 else we'll fail to build, as that file doesn't have knowledge of the new `matcher` field.

### DIFF
--- a/src/ContainerInfo.cpp
+++ b/src/ContainerInfo.cpp
@@ -72,6 +72,14 @@ std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
     return nullptr;
   }
 
+  std::regex matcher;
+  if (std::optional<std::string> str =
+          (*info)["matcher"].value<std::string>()) {
+    matcher = std::regex(*str, std::regex_constants::grep);
+  } else {
+    matcher = std::regex("^" + typeName, std::regex_constants::grep);
+  }
+
   std::optional<size_t> numTemplateParams =
       (*info)["numTemplateParams"].value<size_t>();
 
@@ -122,6 +130,7 @@ std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
 
   return std::unique_ptr<ContainerInfo>(new ContainerInfo{
       std::move(typeName),
+      std::move(matcher),
       numTemplateParams,
       ctype,
       std::move(header),
--- a/src/ContainerInfo.h
+++ b/src/ContainerInfo.h
@@ -16,6 +16,7 @@
 #pragma once
 #include <filesystem>
 #include <optional>
+#include <regex>
 #include <string>
 #include <vector>
 
@@ -73,6 +74,7 @@ const char *containerTypeEnumToStr(ContainerTypeEnum ty);
 
 struct ContainerInfo {
   std::string typeName;
+  std::regex matcher;
   std::optional<size_t> numTemplateParams;
   ContainerTypeEnum ctype = UNKNOWN_TYPE;
   std::string header;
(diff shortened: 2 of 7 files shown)


src/ContainerInfo.cpp
```diff
@@ -72,6 +72,14 @@ std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
     return nullptr;
   }
 
+  std::regex matcher;
+  if (std::optional<std::string> str =
+          (*info)["matcher"].value<std::string>()) {
+    matcher = std::regex(*str, std::regex_constants::grep);
+  } else {
+    matcher = std::regex("^" + typeName, std::regex_constants::grep);
```
jgkamat, line 80: Should we regex escape typeName here?

jgkamat (approved):
Overall, I think this looks good, but I'm a little weary of behavior changing slightly, so I would like to test this on some larger types before landing if you have time. 

Stamping to unblock you, though!

src/ContainerInfo.cpp
JakeHillion: Okay, this ended up being a bit of a rabbit hole:
- According to the comment in my C++ `bits/regex_constants.h` header:
   ```
   /**
   * Specifies that the grammar recognized by the regular expression engine is
   * that used by POSIX utility grep in IEEE Std 1003.1-2001.  This option is
   * identical to syntax_option_type basic, except that newlines are treated
   * as whitespace.
   */
   ```
- According to [IEEE Std 1003.1-2001](https://ieeexplore.ieee.org/stamp/stamp.jsp?tp=&arnumber=974398) (pp. 492):
  ```
  By default, an input line shall be selected if any pattern,
  treated as an entire basic regular expression (BRE) as described in the Base Definitions volume of
  IEEE Std 1003.1-2001, Section 9.3, Basic Regular Expressions, matches any part of the line
  excluding the terminating <newline>;
  ```
- According to section 9.3.3 (BRE Special Characters) (pp. 169), `. [ \ * ^ $` are the special characters. This is obviously not quite correct as the `]` is special in some contexts, but ignoring that for now.

As I understand it (though I'm struggling to find this documented) these won't collide with the C++ names that `typeName` represents, and thus shouldn't need escaping. I'm almost certain our current set are fine without escaping too.

However, if it's easy to escape I'll do it to avoid any unexpected behaviour!

Perhaps the clearer approach would be to deprecate `typeName`, as I think it's only used for this matching, and work towards everything unifying on a regex match for ultimate matching clarity.

JakeHillion:
> Overall, I think this looks good, but I'm a little weary of behavior changing slightly, so I would like to test this on some larger types before landing if you have time.
> 
> Stamping to unblock you, though!

Thanks for the review! I'll absolutely spend some time testing this properly before landing.

src/ContainerInfo.cpp
jgkamat: Ugh, I was expecting there to just be a `std::regex_escape` or similar, but I don't think I see one. In that case I think it should be fine to merge as is (or maybe panic/fail in the case we do see special characters). 

Mainly I don't want this to become a debugging rabbit hole of itself later.

JakeHillion:
Gave this a test and it seems fine with a cache generated from `base.oid.conf`. Landing this for now to get rid of those really ugly `if` conditions - if it causes problems I'll revisit!

PR #12, opened 2022-12-28
move extern repos to https addresses

JakeHillion:
## Summary

The extern repos are currently cloned with SSH. This means that attempting to clone them on a system that doesn't have an SSH key which GitHub accepts fails. Change them to HTTPS as they're read only in this context anyway.

## Test plan

Devserver:
- `cd /tmp && git clone https://github.com/JakeHillion/object-introspection.git oi && cd oi && git checkout https-submodules && ... && make test-devel`

NixOS (no SSH keys):
- `cd /tmp && git clone https://github.com/JakeHillion/object-introspection.git oi && mkdir oi/build && cd oi/build && cmake ..` - fails (unable to clone)
- `cd /tmp && git clone https://github.com/JakeHillion/object-introspection.git oi && mkdir oi/build && cd oi/build && git checkout https-submodules && cmake ..` - successfully clones

&& CI.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,9 +1,9 @@
 [submodule "extern/drgn"]
-	path = extern/drgn
-  url = git@github.com:JakeHillion/drgn.git
+  path = extern/drgn
+  url = https://github.com/JakeHillion/drgn.git
 [submodule "extern/folly"]
-	path = extern/folly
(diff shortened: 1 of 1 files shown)


jgkamat (approved):
LGTM :)

PR #18, opened 2022-12-31
Codegen review changes

tyroguru:
## Summary
Just doing some code review on OICodeGen* and picking up bits I see as I go along. It's mostly minor stuff here and is mainly composed of:

- Removing 'struct' keyword from a bunch of declarations - this is C++ and not C.
- Converting a bunch of dragon error handling cases to use if initializer constructs. Just keeps thing a bit cleaner. Also added in calls to drgn_error_destroy() where missing.
- Some heavy string generation for debug is called in the when debugging is not enabled (glog) so blocked it with a VLOG_IS_ON() macro. 
- Also changed pretty much all postfix increments on for loops. It doesn't actually matter but it's generally accepted to use prefix unless there is an explicit reason and I can't see that there are any in these instances.

## Test plan
CI passes all tests.

### DIFF
--- a/src/OICodeGen.cpp
+++ b/src/OICodeGen.cpp
@@ -85,7 +85,7 @@ OICodeGen::OICodeGen(const Config &c) : config{c} {
   };
 
   config.membersToStub.reserve(typesToStub.size());
-  for (auto &type : typesToStub) {
+  for (const auto &type : typesToStub) {
     config.membersToStub.emplace_back(type, "*");
   }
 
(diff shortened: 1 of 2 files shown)


src/OICodeGen.cpp
```diff
@@ -2976,8 +2968,10 @@ bool OICodeGen::generateJitCode(std::string &code) {
       } else {
         LOG(ERROR) << "Failed to get size: " << err->code << " " << err->message
                    << " " << e;
+        drgn_error_destroy(err);
         return false;
       }
+      drgn_error_destroy(err);
```
JakeHillion, line 2974: Would it be clearer if this was in the positive if branch?

JakeHillion (approved):
few minor comments but looks good

src/OICodeGen.cpp
tyroguru: Actually the way I did it sucks. I've rewritten it with a boolean that is set to true if we should return and then checked below. This allows the error cleanup to be called just once and it's much tidier.

JakeHillion:
looks good :)

PR #23, opened 2023-01-03
Skipped test cleanup

JakeHillion:
## Summary

This change better documents and tracks our skipped tests.

- Add `oil_disable` and `oid_disable` for cases where a test only makes sense for one part of the software. Specifically, any test with `--chase-raw-pointers` can be `oil_disable` as it will never make sense under OIL - it is not a bug like a skipped test.
- Enable void smart pointer present tests.
  - These tests were disabled as they don't track the dynamic size. However, I don't think there is a dynamic size as the void pointer could contain an object of any size and we have no way of knowing it. Therefore, treat the current behaviour of reporting the dynamic size as 0 as correct.
- There are a set of tests in `pointers_incomplete` that codegen under OIL but not OID. Replace the general `skip` with an `oid_skip` to account for this.
- Similarly, top level pointers are skipped over in OID but this isn't an issue in OIL. Replace `skip` with `oid_skip`.
- A field ignore test was `oil_skip`ped as it didn't support it. Enabling the test suggests that it does.
- As OIL gives less rich data, the tests in `pointers_function` pass under it while failing under OID. Enable the OIL tests where appropriate.
- Document every test that remains skipped with a GitHub issue for tracking and discussion, linked to from the code.

Overall, this brings the number of skipped tests down from 63 to 31.

## Test plan

- `make test-devel`
- CI.

### DIFF
--- a/test/integration/multi_arg.toml
+++ b/test/integration/multi_arg.toml
@@ -20,7 +20,7 @@ definitions = '''
 
   # Test that TreeBuilder failing to run on the first arg doesn't impact the second arg
   [cases.tb_fail_first_arg]
-    oil_skip = "oil doesn't handle invalid strings"
+    oil_disable = "treebuilder tests are oid specific"
     param_types = ["const std::string&", "const NodeA&"]
     args = "arg0,arg1"
     setup = """
@@ -32,7 +32,7 @@ definitions = '''
     expect_json = '[{},{"staticSize":12, "dynamicSize":0}]'
 
   [cases.tb_all_fail_crashes]
-    oil_skip = "oil doesn't handle invalid strings"
+    oil_disable = "treebuilder tests are oid specific"
     param_types = ["const std::string&", "const std::string&"]
     args = "arg0,arg1"
     setup = """
--- a/test/integration/pointers.toml
+++ b/test/integration/pointers.toml
@@ -14,7 +14,7 @@ definitions = '''
 
 [cases]
   [cases.int]
-    oid_skip = "top-level pointers are skipped over"
+    oid_skip = "top-level pointers are skipped over" # https://github.com/facebookexperimental/object-introspection/issues/19
     oil_disable = "oil can't chase raw pointers safely"
     param_types = ["int*"]
     setup = "return new int(1);"
@@ -33,7 +33,7 @@ definitions = '''
       ]
     }]'''
   [cases.int_no_follow]
-    oid_skip = "top-level pointers are skipped over"
+    oid_skip = "top-level pointers are skipped over" # https://github.com/facebookexperimental/object-introspection/issues/19
     param_types = ["int*"]
     setup = "return new int(1);"
     expect_json = '''[{
@@ -44,7 +44,7 @@ definitions = '''
       "NOT": "members"
     }]'''
   [cases.int_null]
-    oid_skip = "top-level pointers are skipped over"
+    oid_skip = "top-level pointers are skipped over" # https://github.com/facebookexperimental/object-introspection/issues/19
     param_types = ["int*"]
     setup = "return nullptr;"
     expect_json = '''[{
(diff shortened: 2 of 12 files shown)


test/integration/multi_arg.toml
```diff
@@ -20,7 +20,7 @@ definitions = '''
 
   # Test that TreeBuilder failing to run on the first arg doesn't impact the second arg
   [cases.tb_fail_first_arg]
-    oil_skip = "oil doesn't handle invalid strings"
+    oil_disable = "treebuilder tests are oid specific"
```
ajor, line 23: I think this description is a little misleading. Tried running it and OIL crashes with an exception for this test.

It's more like OIL can't handle invalid data safely, similar to not chasing raw pointers safely, rather than this test not applying to it.

ajor: Another thought: could we wrap OIL's code in a giant try/catch block to handle errors like this?

test/integration/pointers_incomplete.toml
```diff
@@ -42,7 +43,7 @@ definitions = '''
       "NOT": "members"
     }]'''
   [cases.raw_null]
-    skip = "codegen fails on this"
+    oid_skip = "oid codegen fails on this" # https://github.com/facebookexperimental/object-introspection/issues/17
```
ajor, line 46: Don't oid and oil share codegen? How can one work and not the other?

test/integration/pointers.toml
```diff
@@ -69,7 +70,7 @@ definitions = '''
       "NOT": "members"
     }]'''
   [cases.void_no_follow]
-    skip = "top-level pointers are skipped over"
+    oid_skip = "top-level pointers are skipped over" # https://github.com/facebookexperimental/object-introspection/issues/19
```
ajor, line 73: This also appears to be a difference in oid and oil codegen

test/integration/pointers_incomplete.toml
ajor: (I see this is what's happening, just confused!)

JakeHillion: Yep, it's weird. The only bit that I would expect to be different is [the entrypoint generation](https://github.com/facebookexperimental/object-introspection/blob/main/src/OICodeGen.cpp#L3190-L3207), but it doesn't look like the code gets there. Didn't get to the bottom of it so opened the issue for now.

test/integration/multi_arg.toml
JakeHillion: > I think this description is a little misleading. Tried running it and OIL crashes with an exception for this test.
> 
> It's more like OIL can't handle invalid data safely, similar to not chasing raw pointers safely, rather than this test not applying to it.

I think this test is still invalid for oil. I shoehorned oil in for the multi arg tests for compatibility reasons, but the way it runs is:
```
size_t size0 = 0;
auto ret0 = ObjectIntrospection::getObjectSize(&a0, &size0, opts);
std::cout << "{\"ret\": " << ret0 << ", \"size\": " << size0 << "}" << std::endl;
std::cout << "," << std::endl;
size_t size1 = 0;
auto ret1 = ObjectIntrospection::getObjectSize(&a1, &size1, opts);
std::cout << "{\"ret\": " << ret1 << ", \"size\": " << size1 << "}" << std::endl;
std::cout << "]}" << std::endl;
```

As currently, unlike oid, oil has no concept of multi arg probing (except perhaps shoving references in a tuple, but we can't yet do that either). I don't think this test in its current form makes sense for oil, we could instead write different ones to check it doesn't crash.

> Another thought: could we wrap OIL's code in a giant try/catch block to handle errors like this?

This is a test framework error and not an oil error. It tries to add up the static and dynamic size of the `expect_json` to get an oil size, but the expect json for it looks like this:

`expect_json = '[{},{"staticSize":12, "dynamicSize":0}]'`

With no `staticSize` node in the first object. This could be fixed to default to 0 or something, but I think this test is invalid for oil anyway.

ajor (approved):
Thanks for the explanation! Looks good

PR #30, opened 2023-01-18
integration/primitives.toml: Add char8_t test

ajor:
This is a distinct primitive type according to https://en.cppreference.com/w/cpp/language/types
Now all primitive types are tested.

### DIFF
--- a/test/integration/primitives.toml
+++ b/test/integration/primitives.toml
@@ -51,6 +51,10 @@
     param_types = ["wchar_t"]
     setup = "return 'a';"
     expect_json = '[{"staticSize":4, "dynamicSize":0}]'
+  [cases.char8_t]
+    param_types = ["char8_t"]
+    setup = "return 'a';"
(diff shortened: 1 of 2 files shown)


test/integration/primitives.toml
```diff
@@ -51,6 +51,10 @@
     param_types = ["wchar_t"]
     setup = "return 'a';"
     expect_json = '[{"staticSize":4, "dynamicSize":0}]'
+  [cases.char8_t]
```
tyroguru, line 54: Does that complete the list of primitive types as specified by C++20?

tyroguru (approved):

tyroguru (approved):

ajor:
I got too keen with this test - it's not working yet (possibly because the type was only added in C++20)

ajor:
Fixed by enabling char8_t in libclang options

JakeHillion (approved):

Issue #29, opened 2023-01-17
oilgen doesn't set `DRGN_ENABLE_TYPE_ITERATOR=1`


JakeHillion:
This environment variable is required by `oilgen` but isn't automatically set. This provides a surprising error to the end user. It should be set as such to avoid this:

```
if ((setenv("DRGN_ENABLE_TYPE_ITERATOR", "1", 1)) < 0) {
    std::cerr << "Failed to set environment variable\
       DRGN_ENABLE_TYPE_ITERATOR\n";
    exit(EXIT_FAILURE);
  }
```

PR #31, opened 2023-01-19
oilgen: set DRGN_ENABLE_TYPE_ITERATOR

JakeHillion:
## Summary

`oilgen` always uses drgn type iterators but doesn't enable them by itself. Change this so an external environment variable doesn't have to be set.

## Test plan

- `make test-devel`
- CI


tyroguru (approved):

JakeHillion:
closes #29

PR #37, opened 2023-01-25
Polymorphic inheritance

ajor:
Adds support for polymorphic inheritance, gated behind the "--polymorphic-inheritance" command line argument.

This change is split into three commits:
1. Adds tests for polymorphic inheritance. Expected results are based on the old behaviour of not following polymorphic classes.
2. Updates drgn submodule and makes necessary changes to OI code.
3. Adds support for polymorphic inheritance to OI. Updates expected results of polymorphic inheritance tests. The changes in the "test" directory show exactly what differences this new feature makes to the expected results.


src/OICodeGen.cpp
```diff
@@ -986,6 +989,99 @@ std::string OICodeGen::typeToName(drgn_type *type) {
...
+    }
+
+    drgn_type *parent = drgnUnderlyingType(t.type);
+    if (!isDrgnSizeComplete(parent)) {
+      continue;
+    }
+
+    const char *parentName = drgn_type_tag(parent);
+    if (!parentName) {
+      continue;
```
JakeHillion, line 1012: Do these continues warrant some logging?

src/OICodeGen.cpp
```diff
@@ -986,6 +989,99 @@ std::string OICodeGen::typeToName(drgn_type *type) {
...
+ *
+ * drgn only gives us the mapping Class -> Parents, so we must iterate over all
+ * types in the program to build the reverse mapping.
+ */
+bool OICodeGen::enumerateChildClasses() {
+  if (!config.polymorphicInheritance) {
+    return true;
+  }
+
+  if ((setenv("DRGN_ENABLE_TYPE_ITERATOR", "1", 1)) < 0) {
```
JakeHillion, line 1037: Should we remove this? We have to set this env variable mid program in a few places and it doesn't come from upstream. `setenv` called like this can technically only fail with `ENOMEM` so it should be fine either way though.

src/OICodeGen.cpp
```diff
@@ -986,6 +989,99 @@ std::string OICodeGen::typeToName(drgn_type *type) {
   return typeName;
 }
 
+bool OICodeGen::recordChildren(drgn_type *type) {
+  drgn_type_template_parameter *parents = drgn_type_parents(type);
+
+  for (size_t i = 0; i < drgn_type_num_parents(type); i++) {
+    struct drgn_qualified_type t {};
+
+    if (auto *err = drgn_template_parameter_type(&parents[i], &t)) {
```
JakeHillion, line 998: style: I think the new style is to use (explicitly checking the nullptr):
```
if (auto *err = drgn_type_iterator_create(prog, &typesIterator); err != nullptr) {
...
}
```

Though there's nothing enforcing it yet. There are a few instances of this.

src/OICodeGen.cpp
```diff
@@ -1247,6 +1353,41 @@ bool OICodeGen::isEmptyClassOrFunctionType(drgn_type *type,
           drgn_type_num_members(type) == 0);
 }
 
+/*
+ * Returns true iff the provided type
```
JakeHillion, line 1357: Unfinished comment

JakeHillion (approved):
Nice! A few minor comments but looks good. I haven't looked properly at the drgn stuff though so would be good for someone else to cast their eyes over it.

src/OICodeGen.cpp
tyroguru: Yes, I prefer this style and used that in code cleanup for open sourcing.

src/OICodeGen.cpp
```diff
@@ -986,6 +989,99 @@ std::string OICodeGen::typeToName(drgn_type *type) {
...
+      continue;
+    }
+
+    drgn_type *parent = drgnUnderlyingType(t.type);
+    if (!isDrgnSizeComplete(parent)) {
+      continue;
+    }
+
+    const char *parentName = drgn_type_tag(parent);
+    if (!parentName) {
```
tyroguru, line 1011: Again, it's a stylistic thing but I personally much prefer an explicit 'nullptr' check for anything that's not boolean. It's just readability in this cases obviously.

src/OICodeGen.cpp
```diff
@@ -986,6 +989,99 @@ std::string OICodeGen::typeToName(drgn_type *type) {
...
+      continue;
+    }
+
+    // drgn pointers are not stable, so use string representation for reverse
```
tyroguru, line 1015: For multiline comments can we use a C style comment please? Exactly in the style of the one currently at line 1026 below.

src/OICodeGen.cpp
```diff
@@ -1247,6 +1353,41 @@ bool OICodeGen::isEmptyClassOrFunctionType(drgn_type *type,
           drgn_type_num_members(type) == 0);
 }
 
+/*
+ * Returns true iff the provided type
+ */
+bool OICodeGen::isDynamic(drgn_type *type) const {
+  if (!config.polymorphicInheritance) {
```
tyroguru, line 1360: You could group this and the following if together with an || as they both return false in the failure case.

src/OICodeGen.cpp
```diff
@@ -1247,6 +1353,41 @@ bool OICodeGen::isEmptyClassOrFunctionType(drgn_type *type,
...
+  }
+
+  if (drgn_type_virtuality(type) != 0 /*DW_VIRTUALITY_none*/) {
+    // Virtual class - not fully supported by OI yet
+    return true;
+  }
+
+  drgn_type_member_function *functions = drgn_type_functions(type);
+  for (size_t i = 0; i < drgn_type_num_functions(type); i++) {
+    struct drgn_qualified_type t {};
```
tyroguru, line 1375: Just a style thing but I think we tend to use no space before the uniform initializer braces (i.e., t{}). Personally I dislike extra spaces.

tyroguru (commented):
My comments are purely stylistic. Fantastic stuff.

src/OICodeGen.cpp
ajor: Oh yes I remember this one! It's what I was writing when we all got distracted by the wikipedia redesign 😆

src/OICodeGen.cpp
ajor: I don't see the point in this env variable either, but I'll leave it as-is for now to avoid bloating this PR. We might want to change drgn to remove it in the future.

src/OICodeGen.cpp
ajor: I tried changing this but the linter doesn't like it without a space

test/integration/inheritance_polymorphic.toml
```diff
@@ -0,0 +1,120 @@
+# TODO test polymorphic inheritance with classes of the same name in different namespaces
```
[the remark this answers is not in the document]
ajor: I've deleted this comment and opened an issue to remind myself to fix this problem: https://github.com/facebookexperimental/object-introspection/issues/44

src/OICodeGen.cpp
ajor: Ah, it's because I wrote `struct drgn_qualified_type` instead of just `drgn_qualified_type`!

PR #39, opened 2023-01-25
Add CMake hook for defining custom build rules

ajor:
Allows us to use the open source CMakeLists.txt without modification for internal builds with custom environment variables.


JakeHillion (approved):
Nice! Thanks for sorting this out.

PR #40, opened 2023-01-25
Integration tests: Set expected results for cycles.toml

ajor:
Integration tests: Set expected results for cycles.toml

Format the raw_ptr test case with jq to remove tabs and deleted extra data from the JSON.

### DIFF
--- a/test/integration/cycles.toml
+++ b/test/integration/cycles.toml
@@ -1,4 +1,4 @@
-includes = ["memory"]
+includes = ["functional", "memory"]
 definitions = '''
   struct RawNode {
     uint64_t value;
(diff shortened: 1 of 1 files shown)


test/integration/cycles.toml
```diff
@@ -28,88 +28,69 @@ definitions = '''
...
+              }
+            ]
+          }
+        ]
+      }
+    ]
     '''
 
   [cases.unique_ptr]
+    oil_skip = "OIL is not following top-level pointers"
```
ajor, line 93: @JakeHillion This is something that we should be able to support, right?

The test case just has a single raw pointer as the argument to be probed. All other pointers are smart pointers.

JakeHillion: I'm not sure calling it like that makes sense. This (the test framework) makes the argument to OIL `**T`, making the template argument `*T`. It does support (and only supports) following one pointer which you've promised is accurate. Though in hindsight I should probably change that to a reference to be clearer.

JakeHillion (approved):
All looks sensible to me.

test/integration/cycles.toml
JakeHillion: Some more discussion in #43. In short, this isn't something we can blanket enable, particularly with the change to references. We use a lot of incomplete pointers in our tests (`void*` / `class A; A*`) and automatically dereferencing them into a reference globally causes those to all fail. We'd need some extra test configuration to decide whether to dereference a pointer. For now I think it's fine to `oil_skip` or even `oil_disable` any that don't work because of it.

ajor:
I decided in the end to leave the test framework as-is for it's handling of pointer arguments, as I didn't want to disable OIL for most of the other pointer tests.

I worked around the issue in this test by making the argument a reference_wrapper instead of a raw pointer.

PR #41, opened 2023-01-26
Delete old files from test/

ajor:


JakeHillion (approved):
Assuming you've tested this with `make test-devel` and the old Python tests still run. Nice cleanup :)

PR #42, opened 2023-01-26
Oil interface cleanup

JakeHillion:
## Summary

Make a couple of changes to the OIL interface.
- `#ifdef` away the JIT functions which you shouldn't call in compile time mode. They wouldn't work with the header only library (linker failures) but this avoids confusion.
- Change from pointers to references. This is a major change to the API, but they shouldn't have been pointers in the first place as they're always expected to be valid. Bite the bullet now and swap them before we have major consumers.

## Test plan

- CI
- `make test-devel` after each commit
- Manually checked the compile time example after each commit

### DIFF
--- a/examples/compile-time-oil/OilVectorOfStrings.cpp
+++ b/examples/compile-time-oil/OilVectorOfStrings.cpp
@@ -32,7 +32,7 @@ int main() {
   foo.strings.push_back("consectetur adipiscing elit,");
 
   size_t size = -1;
-  int ret = ObjectIntrospection::getObjectSize<Foo>(&foo, &size);
+  int ret = ObjectIntrospection::getObjectSize(foo, size);
 
   std::cout << "oil returned: " << ret << "; with size: " << size << std::endl;
 }
--- a/include/ObjectIntrospection.h
+++ b/include/ObjectIntrospection.h
@@ -79,6 +79,8 @@ enum Response : int {
   OIL_UNINITIALISED = 7,
 };
 
+#ifndef OIL_AOT_COMPILATION
+
 struct options {
   std::string configFilePath{};
   std::string debugFilePath{};
@@ -233,6 +235,8 @@ int getObjectSize(T *ObjectAddr, size_t *ObjectSize, const options &opts,
   return CodegenHandler<T>::getObjectSize(ObjectAddr, ObjectSize, opts,
                                           checkOptions);
 }
+
+#endif
 
 /*
  * You may only call this after a call to the previous signature, or a
--- a/include/ObjectIntrospection.h
+++ b/include/ObjectIntrospection.h
@@ -118,7 +118,7 @@ class OILibrary {
  private:
   class OILibraryImpl *pimpl_;
 
-  size_t (*fp)(void *) = nullptr;
+  size_t (*fp)(const void *) = nullptr;
 };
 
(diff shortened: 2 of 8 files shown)


include/ObjectIntrospection.h
```diff
@@ -139,25 +141,25 @@ class CodegenHandler {
     delete lib;
   }
 
-  static int getObjectSize(T *ObjectAddr, size_t *ObjectSize) {
+  static int getObjectSize(T &ObjectAddr, size_t &ObjectSize) {
```
ajor, line 144: Can we make these ObjectAddrs const references? (and the same for the other functions)

```suggestion
  static int getObjectSize(const T &ObjectAddr, size_t &ObjectSize) {
```

examples/compile-time-oil/OilVectorOfStrings.cpp
```diff
@@ -32,7 +32,7 @@ int main() {
   foo.strings.push_back("consectetur adipiscing elit,");
 
   size_t size = -1;
-  int ret = ObjectIntrospection::getObjectSize<Foo>(&foo, &size);
+  int ret = ObjectIntrospection::getObjectSize(foo, size);
```
ajor, line 35: I'm not sure what the right thing to do is here, just pointing it out for discussion...

Making the size argument a reference instead of a pointer means it's less clear to users of the API that it will be modified

ajor (commented):
Go references! 🏆 

We'll also want to update the big block comment at the top of ObjectIntrospection.h to match the new API.

examples/compile-time-oil/OilVectorOfStrings.cpp
JakeHillion: Agreed. I wasn't sure about this change. On the one hand, a non-const reference is more accurate as the API requires an always valid pointer to something which it then writes to, and adding a distinct error condition for giving it a null pointer seems pointless. On the other hand, non-const references for read only stuff are kind of ingrained. However, if it should be changed it should definitely be done now with the `T &` change.

include/ObjectIntrospection.h
JakeHillion: It seems like we can. New commit added to this diff doing that.

examples/compile-time-oil/OilVectorOfStrings.cpp
JakeHillion: 🦀 - `get_object_size(&obj, &mut size);` solves this nicely, though you wouldn't be using a return arg there.

ot: In C++ it is more idiomatic to accept a mutable reference. If you use a pointer, you need to deal with `nullptr`.

But it is even more idiomatic to *return* the value. You can use `folly::Expected` or something similar, which is more self-documenting. Or an exception.

JakeHillion: Yes, I think if `std::expected` was in I'd be using that. Unfortunately exceptions are sort of out as both the `OIL_UNINITIALISED` result and the `OIL_SUCCESS` result need to be on the fast path in different situations, and exceptions would make `OIL_SUCCESS` faster. Thanks for the input!

ajor (approved):

PR #43, opened 2023-01-26
OIL gentests top level pointers

JakeHillion:
## Summary

Allow top level pointers in the generated OIL tests. This is a bit of an RFC as it has some shortcomings. Notably this is much harder to do with the references - because both paths of the constexpr must compile unless it's in a generic function it had to be extracted, whereas in the previous implementation an additional pointer layer was fine but wouldn't return much.

As the CI should show this breaks as many tests as it fixes. The change from references to pointers (*&) to references (&) breaks the previous compatibility with no follow tests as the top level reference is deliberately ignored, whereas the previous (** / *&) included one.

## Test plan

It's not.

### DIFF
--- a/examples/compile-time-oil/OilVectorOfStrings.cpp
+++ b/examples/compile-time-oil/OilVectorOfStrings.cpp
@@ -32,7 +32,7 @@ int main() {
   foo.strings.push_back("consectetur adipiscing elit,");
 
   size_t size = -1;
-  int ret = ObjectIntrospection::getObjectSize<Foo>(&foo, &size);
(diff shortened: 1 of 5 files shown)


JakeHillion:
Leaving this around for posterity, but I'll close it as it won't be landed in its current form. While it is possible to dereference pointers statically in OIL tests, it doesn't make sense on `void`/incomplete pointers. Perhaps we could add a flag to the test configs for dereferencing in OIL, but we can't do it blanket as it causes as many failures as it solves.

PR #46, opened 2023-01-31
oil: switch to using an impl function with no default implementation

JakeHillion:
## Summary

Although the existing weak symbol implementation works well in the context it was tested in, it doesn't work well in the static library context. Switch to an impl weak symbol which has no default implementation and check that in the library.

## Test plan

- CI
- `cd examples/compile-time-oil/ && make && ./OilVectorOfStrings`

### DIFF
--- a/include/ObjectIntrospection.h
+++ b/include/ObjectIntrospection.h
@@ -111,7 +111,7 @@ class OILibrary {
   OILibrary(void *TemplateFunc, options opt);
   ~OILibrary();
   int init();
-  int getObjectSize(void *ObjectAddr, size_t &size);
+  int getObjectSize(void *objectAddr, size_t &size);
 
   options opts;
 
@@ -141,17 +141,17 @@ class CodegenHandler {
     delete lib;
   }
 
-  static int getObjectSize(const T &ObjectAddr, size_t &ObjectSize) {
+  static int getObjectSize(const T &objectAddr, size_t &objectSize) {
     OILibrary *lib;
     if (int responseCode = getLibrary(lib);
         responseCode != Response::OIL_SUCCESS) {
       return responseCode;
     }
 
-    return lib->getObjectSize((void *)&ObjectAddr, ObjectSize);
+    return lib->getObjectSize((void *)&objectAddr, objectSize);
   }
 
-  static int getObjectSize(const T &ObjectAddr, size_t &ObjectSize,
+  static int getObjectSize(const T &objectAddr, size_t &objectSize,
                            const options &opts, bool checkOptions = true) {
     OILibrary *lib;
     if (int responseCode = getLibrary(lib, opts, checkOptions);
@@ -159,7 +159,7 @@ class CodegenHandler {
       return responseCode;
     }
 
-    return lib->getObjectSize((void *)&ObjectAddr, ObjectSize);
+    return lib->getObjectSize((void *)&objectAddr, objectSize);
   }
 
  private:
@@ -230,11 +230,17 @@ class CodegenHandler {
  * Ahead-Of-Time (AOT) compilation.
  */
 template <class T>
-int getObjectSize(const T &ObjectAddr, size_t &ObjectSize, const options &opts,
+int getObjectSize(const T &objectAddr, size_t &objectSize, const options &opts,
                   bool checkOptions = true) {
-  return CodegenHandler<T>::getObjectSize(ObjectAddr, ObjectSize, opts,
+  return CodegenHandler<T>::getObjectSize(objectAddr, objectSize, opts,
                                           checkOptions);
 }
+
+#else
+
+template <class T>
+int __attribute__((weak))
+getObjectSizeImpl(const T &objectAddr, size_t &objectSize);
 
 #endif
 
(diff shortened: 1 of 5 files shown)


jgkamat (approved):
Looks good to me! :)

src/OIGenerator.cpp
```diff
@@ -76,8 +76,20 @@ OIGenerator::findOilTypesAndNames(drgnplusplus::program& prog) {
       linkageName = linkageNameCstr;
     }
 
+    const std::string_view linkagePrefix =
+        "_ZN19ObjectIntrospection13getObjectSize";
```
jgkamat, line 80: This isn't a stable name - so I'm not sure if we can depend on it in the long term - but I think it's fine for now.

include/ObjectIntrospection.h
```diff
@@ -248,10 +254,12 @@ int getObjectSize(const T &ObjectAddr, size_t &ObjectSize, const options &opts,
  * production system.
  */
 template <class T>
-int __attribute__((weak))
-getObjectSize(const T &ObjectAddr, size_t &ObjectSize) {
+int getObjectSize(const T &ObjectAddr, size_t &ObjectSize) {
```
ot, line 257: Independently from this diff, there is some inconsistency about code style, the `&` should go with the type, not with the variable. Also, arguments should be in mixedCase.

src/OIGenerator.cpp
```diff
@@ -76,8 +76,20 @@ OIGenerator::findOilTypesAndNames(drgnplusplus::program& prog) {
...
+    if (!linkageName.starts_with(linkagePrefix)) {
+      LOG(INFO) << "expected symbol to match but missing linkagePrefix: `"
+                << linkageName << "`";
+      continue;
+    }
```
ot, line 85: Do we need to support both for backward compatibility? Should we have a TODO at some point to remove this?

src/OIGenerator.cpp
```diff
@@ -76,8 +76,20 @@ OIGenerator::findOilTypesAndNames(drgnplusplus::program& prog) {
...
     LOG(INFO) << "found linkage name: " << linkageName;
-    out.push_back({paramType, linkageName});
+
+    std::string weakLinkageName =
+        "_ZN19ObjectIntrospection17getObjectSizeImpl" +
```
ot, line 89: Do we hardcode mangled names elsewhere?

src/OIGenerator.cpp
```diff
@@ -76,8 +76,20 @@ OIGenerator::findOilTypesAndNames(drgnplusplus::program& prog) {
...
+        linkageName.substr(linkagePrefix.length());
+    LOG(INFO) << "made weak linkage name: " << weakLinkageName;
```
ot, line 91: Do we need to keep this log?

include/ObjectIntrospection.h
JakeHillion: I don't think the former is inconsistent in OI, if you look at say `src/OICodeGen.cpp` which was manually formatted the qualifiers are attached to the name instead of the type. I can change the name of the variables though, that is inconsistent.

src/OIGenerator.cpp
JakeHillion: No, fortunately this code isn't widely used yet so no backwards compatibility is fine.

src/OIGenerator.cpp
JakeHillion: > This isn't a stable name - so I'm not sure if we can depend on it in the long term - but I think it's fine for now.
@jgkamat 

Yes, this isn't ideal. The `ObjectIntrospection::getObjectSizeImpl` doesn't appear in the DWARF due to not having a body. I kind of think this is a bug, as the types are perfectly available, but it's a bug/feature that's largely out of our control. I guess the idea is that the strong symbol's debug would collide with the weak symbol's, and be more informative?

> Do we hardcode mangled names elsewhere?
@ot 

Not commonly, though there is one instance in [OICompiler.cpp:341](https://github.com/facebookexperimental/object-introspection/blob/23026d80ba9ac85409a3959f010eca4788737f0a/src/OICompiler.cpp#L341-L342).

The issue I'm concerned about with demangling this name is I'd then have to remangle it to get from the top one to the impl. If the impl was in the DWARF I could find it from there, but it isn't. Perhaps it would be more generic if I did `s/13getObjectSize/17getObjectSizeImpl/` but that also feels sketchy. Open to suggestions!

JakeHillion: Since writing that I've changed my mind, going to attempt to work with the demangled names and keep a reverse mapping.

src/OIGenerator.cpp
```diff
@@ -18,40 +18,77 @@
...
+  static constexpr std::string_view strongSymbolPrefix =
+      "int ObjectIntrospection::getObjectSize<";
+  static constexpr std::string_view weakSymbolPrefix =
+      "int ObjectIntrospection::getObjectSizeImpl<";
+
+  std::unordered_map<std::string, std::pair<std::string, std::string>>
+      templateArgsToSymbolsMap;
+
+  auto symbols = prog.find_symbols_by_name(nullptr);
+  for (drgn_symbol* sym : *symbols) {
```
jgkamat, line 42: I would just double check this dosen't slow things down too much. I don't think it should though!

src/OIGenerator.cpp
```diff
@@ -18,40 +18,77 @@
...
+    auto symName = drgnplusplus::symbol::name(sym);
+    auto demangled = boost::core::demangle(symName);
+
+    if (demangled.starts_with(strongSymbolPrefix)) {
```
jgkamat, line 46: This is unlikely for our case since we control the sites, but I wanted to flag the behavior here in the case where multiple mangled names map to one demangled name. I think the last one in the symbol table is selected - is that alright?

It might be safer to at least log a warning in that case so we know if something fishy is happening.

jgkamat (approved):
Overall looks good to me! Made a couple notes to just flag some code, but things looks solid :)

src/OIGenerator.cpp
JakeHillion: In terms of the dereference? Looked at the ASM and it's fine. If it's about whether that drgn function loads all the DWARF, that's a problem I'm happy to push down the road for now - this program is slow until I add the new drgn feature anyway. Thanks for the check!

PR #48, opened 2023-01-31
config: look up paths relative to config file

JakeHillion:
## Summary

Looks up the paths in the config file relative to it. This should mean that although we get relative paths from our internal build system (buck2) we can still use the config file.

## Test plan

- `make test-devel` (absolute paths)
- CI (relative paths)
- test internally to check it actually does let us do that


ajor (approved):
Looks good. As discussed, the behaviour of filesystem::path::append for absolute paths isn't obvious so could do with a comment

PR #49, opened 2023-02-01
CI: Split build and test into separate jobs

ajor:
This will give us more flexibility later, e.g. allowing different steps for GCC and Clang (as needed for code coverage), or letting us parallelise our tests across multiple machines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,22 +1,41 @@
 version: 2.1
 
 workflows:
-  build-test:
+  object-introspection:
     jobs:
       - lint
-      - build-test:
-          name: build-test-gcc
+
+      - build:
+          name: build-gcc
           cc: /usr/bin/gcc
           cxx: /usr/bin/g++
-      - build-test:
-          name: build-test-clang
+      - test:
+          name: test-gcc
+          requires:
+            - build-gcc
+
+      - build:
+          name: build-clang
           cc: /usr/bin/clang-12
           cxx: /usr/bin/clang++-12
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -1,22 +1,41 @@
...
+            - build-clang
 
-jobs:
-  lint:
+executors:
+  ubuntu-docker:
     docker:
       - image: ubuntu:jammy
+    resource_class: small
+  big-boy:
```
JakeHillion, line 31: Sticking with this are we? I like it.

.circleci/config.yml
```diff
@@ -42,10 +61,9 @@ jobs:
             black --check --diff test/
             isort --check --diff test/
 
-  build-test:
-    machine:
-      image: ubuntu-2204:2022.10.2
-    resource_class: 2xlarge
+  build:
+    # TODO this job could be run in Docker
```
JakeHillion, line 65: leaving this for now?

JakeHillion (approved):
Nice! Especially nice for retrying only the tests when they fail, with or without SSH.

.circleci/config.yml
ajor: Yeah, I've had enough of working on the CI so did plan on leaving this for now.

There are a few differences between the docker and machine Ubuntu images that'll take a little bit of time to work through.
Also, since we're building oid as a dynamically linked executable, we'd need to be certain we have the same shared libraries on the build and test jobs.

PR #50, opened 2023-02-01
CI: Output junit result format directly from CTest

ajor:
Remove the conversion script we were previously using.


JakeHillion (approved):
approving commit 2

PR #51, opened 2023-02-01
Parallel tests

ajor:
CI: Parallelise tests
    
    This reduces the time to run tests from 5-9 minutes down to 3-4 minutes.
    
    `nproc` is returning `1` on the CI node which is processing our
    config.yml. In the absense of a better way of getting the number of CPUs
    on our executor, just hardcode it.
    
    We actually have 16 cores on our 2xlarge executors, but higher
    parallelism in tests appears to cause oid to sometimes fail with SIGILL.
    I'm still investigating this problem.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -141,9 +141,8 @@ jobs:
             OMP_NUM_THREADS: 1
           command: |
             echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
-            ctest --test-dir build/test/ --test-action Test -j$(nproc) \
+            ctest --test-dir build/test/ --test-action Test -j 4 \
               --no-compress-output --output-on-failure \
-              --exclude-regex 'TestTypes\/ComparativeTest\..*' \
               --schedule-random --timeout 30 --repeat until-pass:2 \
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -86,38 +102,48 @@ jobs:
...
+              llvm-12-dev
       - run:
           name: Test
           environment:
             # disable drgn multithreading as tests are already run in parallel
             OMP_NUM_THREADS: 1
           command: |
             echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
-            cp test/ci.oid.toml build/testing.oid.toml
-            ctest --test-dir build/test/ --test-action Test -j$(nproc) \
```
JakeHillion, line 110: Did you perhaps try `$$(nproc)`? I haven't seen anything suggesting it would work but if the issue is it being substituted in two places it might. Hardcoding is definitely fine though.

ajor: I've updated the PR's summary with an explanation about the number of cores to use. We can't actually use `nproc` because of some bugs with too high parallelism. 😢

JakeHillion (approved):

Issue #62, opened 2023-02-08
Write test for folly::sorted_vector_set


tyroguru:
As per the description, we need a test writing for folly::sorted_vector_set.

JakeHillion:
Closed with #80

PR #63, opened 2023-02-08
Fix folly::sorted_vector_set and minor codegen comment change

tyroguru:
## Summary

This change corrects how we handle folly::sorted_vector_set handling and adds another space into the codegen comments (so you can easily cut'n'paste the symbol without picking up the trailing slash).

I have logged task #62 as this change doesn't contain a test for this container and we need one.

## Test plan
'make test-devel` passes all tests. The integration tests are failing for me currently but it's not owing to anything here (I think!).


JakeHillion (approved):
Double checking that this works, as the test plan only says run `make test-devel` while acknowledging there's no test for it. Assuming it works, the container change lgtm.

tyroguru:
Yeah, I've tested these changes against an application and have just transplanted the changes across.

PR #65, opened 2023-02-10
Remove type path printing debug code 

tyroguru:
## Summary
When verbose logging is enabled (level 2 and above) we generate a huge amount of debug and a lot of it is coming from the printTypePath() function which is called when we enumerate types. This code iterates over all the known types and prints the name out which generates a lot of noise and makes it more difficult to figure out what is going off.

I've spoken to @arsarwade and he is going to rewrite the debugging code in codegen at some point. In the meantime I'm just removing this before I lose my mind.

## Test plan
`make test-devel`works with no new failures.


JakeHillion (approved):

JakeHillion:
There are 3 commits here, is this intentional?

PR #67, opened 2023-02-10
Fix/remove some log statements

arsarwade:
Summary:
Remove some more log statements to reduce output

Test Plan:
make test-devel

### DIFF
--- a/src/OICodeGen.cpp
+++ b/src/OICodeGen.cpp
@@ -537,7 +537,6 @@ bool OICodeGen::buildNameInt(drgn_type *type, std::string &nameWithoutTemplate,
       return false;
     }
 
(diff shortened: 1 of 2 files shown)


src/OICodeGen.cpp
```diff
@@ -3540,11 +3525,9 @@ bool OICodeGen::generateNamesForTypes() {
     }
 
     std::string tname = templateTransformType(name);
-    VLOG(1) << "BOO type " << e << " " << name;
```
ajor, line 3543: 😆

ajor (approved):
First PR the code coverage bot has commented on! 🎉

The reduced coverage is nothing to worry about here though, obviously.

PR #75, opened 2023-02-20
Special case conditional_t type name handling

tyroguru:
## Summary
A '`conditional_t` type's name will often contain conditional expressions so we need to special case this as it causes problems when removing namespaces.

## Test plan
This change fixes issues with volume testing that were being encountered. Current tests have no new failures. A new test needs to be added to include `conditional_t` types (see issue #76 ).

### DIFF
--- a/src/OICodeGen.cpp
+++ b/src/OICodeGen.cpp
@@ -263,6 +263,14 @@ std::string OICodeGen::stripFullyQualifiedNameWithSeparators(
   std::vector<std::string> stack;
   std::string sep = " ,<>()";
   std::string tmp;
+  constexpr std::string_view cond = "conditional_t";
+  constexpr std::string_view stdCond = "std::conditional_t";
+  static int cond_t_val = 0;
+
+  if ((fullyQualifiedName.starts_with(cond)) ||
+      (fullyQualifiedName.starts_with(stdCond))) {
+    return "conditional_t_" + std::to_string(cond_t_val++);
+  }
 
   for (auto &c : fullyQualifiedName) {
     if (sep.find(c) == std::string::npos) {


src/OICodeGen.cpp
```diff
@@ -263,6 +263,17 @@ std::string OICodeGen::stripFullyQualifiedNameWithSeparators(
   std::vector<std::string> stack;
   std::string sep = " ,<>()";
   std::string tmp;
+  std::string cond = "conditional_t";
```
JakeHillion, line 266: `s/std::string/std::string_view/`

src/OICodeGen.cpp
```diff
@@ -263,6 +263,17 @@ std::string OICodeGen::stripFullyQualifiedNameWithSeparators(
...
+  static int cond_t_val = 0;
+
+  /*
+   * A "conditional_t/std::conditional_t" has to be special cased as its name
+   * may contain conditional statements.
+   */
+  if ((fullyQualifiedName.find(cond, 0) == 0) ||
```
JakeHillion, line 273: `fullyQualifiedName.find(` -> `fullyQualifiedName.starts_with(`

src/OICodeGen.cpp
```diff
@@ -263,6 +263,17 @@ std::string OICodeGen::stripFullyQualifiedNameWithSeparators(
...
+      (fullyQualifiedName.find(cond, 5) == 5)) {
```
JakeHillion, line 274: `fullyQualifiedName.find(` -> `fullyQualifiedName.starts_with("std::" + cond)`

JakeHillion (approved):

JakeHillion:
😎

PR #77, opened 2023-02-21
Add integration tests for basic C++ language features

ajor:
Now covered:
- simple classes and unions
- enums
- typedefs
- arrays
- class templates

These also form the basic of testing DWARF parsing in type-graph


ajor:
Looks like we might have come across a difference between GCC and Clang!

I'll need to investigate this test failure.

ajor:
The error comes from drgn: `DW_AT_upper_bound is too large`

GCC appears to produce this DWARF for a zero-length array:
```
 <1><298874>: Abbrev Number: 125 (DW_TAG_array_type)                                           
    <298875>   DW_AT_type        : <0x14d>                                                     
    <298879>   DW_AT_sibling     : <0x29888c>                                                  
 <2><29887d>: Abbrev Number: 347 (DW_TAG_subrange_type)                                        
    <29887f>   DW_AT_type        : <0x37>
    <298883>   DW_AT_upper_bound : 0xffffffffffffffff
```

jgkamat:
This is because (for some reason) the format for gcc zero length arrays changed slightly.
https://github.com/osandov/drgn/commit/12b0214b4d7fb6834a3e0a92e0d2a634abe8e8e4
This is the support that was written so far, but it assumes `DW_FORM_sdata`
But now this is :
```
 [   b16]    array_type           abbrev: 69
             type                 (ref4) [    5a]
             sibling              (ref4) [   b2d]
 [   b1f]      subrange_type        abbrev: 70
               type                 (ref4) [    2e]
               upper_bound          (data8) 18446744073709551615
```

a data8. I'll have to see if we can just try to read a generic integer with elfutils instead of hardcoding sdata there.

jgkamat:
It seems like this is a discrepancy between c and c++ - if gcc compiles c code the output is readable by drgn, otherwise it results in the above.

test/integration/gen_tests.py
```diff
@@ -61,6 +61,23 @@ def add_test_setup(f, config):
...
     # fmt: on
 
+    def get_param_str(param, i):
+        if "]" in param:
+            # Array param
+
+            if ")" in param:
+                # "int(&)[5]" ->  "int (&a0)[5]"
+                start, end = param.split(")")
+                return f"{start}a0){end}"
```
ajor, line 71: Don't hardcode "a0"

test/integration/gen_tests.py
```diff
@@ -61,6 +61,23 @@ def add_test_setup(f, config):
...
+
+            # "int[5]" -> "int a0[5]"
+            # "int[5][10]" -> "int a0[5][10]"
+            type_name, array_size = param.split("[", 1)
+            return f"{type_name} a0[{array_size}"
```
ajor, line 76: Don't hardcode "a0"

test/integration/arrays.toml
```diff
@@ -0,0 +1,77 @@
...
+        "staticSize":40,
+        "dynamicSize":0,
+        "length":10,
+        "capacity":10,
+        "elementStaticSize":4
+      }]}]'''
+  [cases.member_int0]
+    # WARNING: zero-length arrays are handled differently to non-empty arrays.
+    # They end up not being treated as containers. This should probably change
+    # in the future.
```
ajor, line 29: Looks like `TreeBuilder::isContainer` might be responsible for this behaviour

ajor:
Updated drgn to pull in Jay's zero-length array fix: https://github.com/osandov/drgn/pull/279
I put this commit at the bottom of our stack of custom commits to hopefully make rebasing on upstream easier in the future.

jgkamat (approved):
Thanks so much for adding more tests! :)

Issue #59, opened 2023-02-06
SIGILL in integration_test_target during highly parallel test runs


ajor:
This can be reliably reproduced on the CI by upping the parallelism from 4 to 16.

The target process receives a SIGILL, with the instruction pointer pointing to the `dataBase` address inserted by OID. The "illegal instruction" is the pointer which is stored at this address.

Info extracted from a core dump of the target process:
```
(gdb) x/i $rip
=> 0x7fc01aefe040:      (bad)  
(gdb) p $rip
$1 = (void (*)(void)) 0x7fc01aefe040
(gdb) x/xg $rip
0x7fc01aefe040: 0x00007fc01adfe027
```

In order to get a core dump from the target process, oid needs to be modified to not catch SIGILLs - this case needs commenting out: https://github.com/facebookexperimental/object-introspection/blob/72f31639b7c1770fe9e5d5c908e1db79d0a5ddba/src/OIDebugger.cpp#L1255


Extracts from OID log:
```
I0202 15:46:23.986310 10953 OICompiler.cpp:595] Relocated code 0x7fba5281b000 to 7fc01aefe080                                                                         
I0202 15:46:23.986321 10953 OICompiler.cpp:604] Relocated data 0x7fba5281b400 to 7fc01aefe480                                                                         
I0202 15:46:23.986332 10953 OICompiler.cpp:604] Relocated data 0x7fba5291b420 to 7fc01affe4a0                                                                         
I0202 15:46:23.986343 10953 OICompiler.cpp:604] Relocated data 0x7fba5291b431 to 7fc01affe4b1                                                                                                                                                                                                                                                
I0202 15:46:23.986353 10953 OICompiler.cpp:604] Relocated data 0x7fba5291b460 to 7fc01affe4e0                                                                         
I0202 15:46:23.986377 10953 OICompiler.cpp:331] findSymbol(cookieValue) = Synth 7fc01aefe050                                                                          
I0202 15:46:23.986392 10953 OICompiler.cpp:331] findSymbol(dataBase) = Synth 7fc01aefe040                                                                             
I0202 15:46:23.986403 10953 OICompiler.cpp:331] findSymbol(dataSize) = Synth 7fc01aefe048                                                                             
I0202 15:46:23.986420 10953 OICompiler.cpp:331] findSymbol(logFile) = Synth 7fc01aefe058
...
[...]

PR #85, opened 2023-02-28
reduce race window for breakpoint trap in prologue

tyroguru:
## Summary
The CI has been experiencing failed tests owing to targets dying with SIGILL delivery and as far as I know we've never seen this in production which always seemed odd. The reason is that we have a race condition in processTraps() that becomes more reproducible as the JIT code get smaller (quicker to execute).

The race is between when we continue a target thread to execute the JIT code (the ptrace(PTRACE_CONT) at the bottom of processTraps()) and waiting for that thread to finish JIT code execution - to execute the int3 instruction and the subsequent SIGTRAP that is raised to oid which we receive by the waitpid() call at the start of processTrap().

The VLOG() calls being executed at the start of processTraps() and the ptrace(PTRACE_CONT) give a window where we are not waiting for signals delivered to the target thread. Now, in theory, this shouldn't matter as when a thread is controlled by ptrace (any previously seized threads in our case) any and all signals delivered to it should be redirected to the tracer (i.e., oid). The oid process should therefore receive a SIGTRAP for the int3 instruction in the prologue code when it next executes a waitpid() call. However, the symptoms were that we received a SIGILL from executing junk after the prologue section - it appeared as though we never executed the int3 instruction! However, if I replaced the int3 with ud2 then we do execute it and subsequent bpftrace'ing shows that the int3 is dispatched in the kernel. Why oid never appears to receive it is a mystery to me at the minute and it's either a bug in a ptrace related area, how we pick signals up or in ptrace itself. Further investigation needs to be done.

What this change does is to remove the ptrace(PTRACE_CONT) at the start of processTrap() as I don't think we actually need it (see below). All this does is remove the cycles that exist between oid continuing the target and waiting for a signal. I think we have much work to do to close this race and it's still pretty wide open but we get away with it even with these changes because our JIT code is so large! In my tests it appears that the simplest possible case of an class with a single integer member still generates ~800 instructions. more work to be done there also.

I also introduced a new VLOG() level of 4 which is intended now for al logging calls that are executed while a thread is stopped (i.e., really, really don't execute these logs unless we absolutely need to).

## Test plan

The CI now passes all tests without any repeated tests being needed so I have changed that in the config. The thing I'm most concerned about is not the second call to processTraps() when we are cleaning threads up and releasing any that have received/generated signals or extended wait events while other things were happening. I've done some mental gymnastics and convinced myself that we are OK . I've also been experimenting with some pretty aggressive multithreaded test cases which have been OK. However, the real test comes with very aggressive hugely threaded applications and I have yet to do that testing.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -149,9 +149,9 @@ jobs:
             OMP_NUM_THREADS: 1
           command: |
             echo 0 | sudo tee /proc/sys/kernel/yama/ptrace_scope
-            ctest --test-dir build/test/ --test-action Test -j 4 \
+            ctest --test-dir build/test/ --test-action Test -j 16 \
               --no-compress-output --output-on-failure \
-              --schedule-random --timeout 30 --repeat until-pass:4 \
+              --schedule-random --timeout 60 \
               --output-junit results.xml
       - store_test_results:
           path: build/test/results.xml
--- a/src/OIDebugger.cpp
+++ b/src/OIDebugger.cpp
@@ -439,7 +439,7 @@ bool OIDebugger::isExtendedWait(int status) {
 }
 
 bool OIDebugger::contTargetThread(bool detach) const {
-  VLOG(1) << "contTargetThread: About to PTRACE_CONT pid " << std::dec
+  VLOG(4) << "contTargetThread: About to PTRACE_CONT pid " << std::dec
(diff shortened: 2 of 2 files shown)


JakeHillion:
closes #59

JakeHillion (approved):
Trusting there's been some good testing of the ptrace cont being removed.

PR #86, opened 2023-02-28
tests:  make relative config paths absolute when copying

### LINKED ISSUE #55: Test cases with a `config` section may fail with a relative config
The config section causes a new config to be created in a different temporary location (https://github.com/facebookexperimental/object-introspection/blob/5c5bfd315c4095c0cfbb869576040a0a9719bca7/test/integration/runner_common.cpp#L155-L163). Now that the config file can contain relative paths this may break it.

Possibly the easiest solution to this is to canonicalise the paths in the base config first, allowing it to be moved anywhere after.

JakeHillion:
## Summary

As config files may contain paths relative to them, copying them into an arbitrary location (tmp path) won't succeed. Due to the directory layout this works in OSS & CI, but it failed when the tests were brought over internally. Canonicalise the paths while copying the config file to prevent this issue.

It became clear while testing this that the tests in `ignored.toml` weren't being run. Added them to the `test/integration/CMakeLists.txt` test cases and fixed them working with OIL.

Closes #55 

## Test plan

- D43647536
- `make test-devel`
- CI

### DIFF
--- a/test/integration/runner_common.cpp
+++ b/test/integration/runner_common.cpp
@@ -1,5 +1,8 @@
 #include "runner_common.h"
 
+#include <toml++/toml.h>
+
+#include <algorithm>
 #include <boost/algorithm/string.hpp>
 #include <boost/asio.hpp>
 #include <boost/process.hpp>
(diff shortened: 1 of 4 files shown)


test/integration/runner_common.h
```diff
@@ -10,6 +10,8 @@
 #include <iostream>
 #include <string>
 
+namespace fs = std::filesystem;
```
ajor, line 13: It's generally bad practice to have using declarations or namespace aliases in a header file as they then pollute the namespaces of anything that includes the header.

Things are normally fully-qualified in the header file and then we define and use the namespace aliases in the CPP file instead.

test/integration/runner_common.cpp
```diff
@@ -119,6 +121,48 @@ int IntegrationBase::exit_code(Proc &proc) {
   return proc.proc.exit_code();
 }
 
+fs::path IntegrationBase::createCustomConfig(std::string extraConfig) {
+  boost::trim(extraConfig);
```
ajor, line 125: Is toml++ not ok with whitespace?

test/integration/runner_common.cpp
```diff
@@ -119,6 +121,48 @@ int IntegrationBase::exit_code(Proc &proc) {
   return proc.proc.exit_code();
 }
 
+fs::path IntegrationBase::createCustomConfig(std::string extraConfig) {
```
ajor, line 124:
```suggestion
fs::path IntegrationBase::createCustomConfig(std::string_view extraConfig) {
```
(or `const std::string &`)
If we don't need a copy of the string (depends on the boost trim thing)

test/integration/runner_common.cpp
JakeHillion: The reason this was initially implemented was added whitespace when providing the extra config. I resolved this so this was removed.

ajor (approved):

PR #90, opened 2023-03-03
Add support for exclusive size reporting

jgkamat:
## Summary
Add initial pass for exclusive size

Unfortunately this became a lot more complex than I was hoping - I was trying to avoid making additional calls into rocksdb, so I couldn't just query for members. That said, it's not that bad considering I had to touch a lot of places.

## Test plan
Writing tests to hopefully cover all the edge cases I can think of.

### DIFF
--- a/src/TreeBuilder.cpp
+++ b/src/TreeBuilder.cpp
@@ -156,9 +156,15 @@ struct TreeBuilder::Node {
 
   std::optional<bool> isset{std::nullopt};
 
+  /**
+   * An estimation of the "exclusive size" of a type, trying to
+   * attribute every byte once and only once to types in the tree.
+   */
+  size_t exclusiveSize{};
+
   MSGPACK_DEFINE_ARRAY(id, name, typeName, typePath, isTypedef, staticSize,
                        dynamicSize, paddingSavingsSize, containerStats, pointer,
-                       children, isset)
+                       children, isset, exclusiveSize)
 };
 
 TreeBuilder::~TreeBuilder() {
@@ -376,6 +382,19 @@ static std::string_view drgnKindStr(struct drgn_type *type) {
   return kind;
 }
 
+void TreeBuilder::setSize(TreeBuilder::Node &node, uint64_t dynamicSize,
+                          uint64_t memberSizes) {
+  node.dynamicSize = dynamicSize;
+  if (memberSizes > node.staticSize + node.dynamicSize) {
+    // TODO handle this edge case in a more elegant way.
+    // This can occur when handling bitfields or vector<bool>
+    // (as we cannot accurately track sub-byte sizes currently)
+    node.exclusiveSize = 0;
+  } else {
+    node.exclusiveSize = node.staticSize + node.dynamicSize - memberSizes;
+  }
+}
+
 TreeBuilder::Node TreeBuilder::process(NodeID id, Variable variable) {
   Node node{
       .id = id,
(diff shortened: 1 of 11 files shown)


JakeHillion (approved):
Cool. Slightly melted my brain, but i think the logic is sound. Can you add exclusive size to the `std_vector.toml/cases.reserve` test case? My understanding is the unused size in the vector will be added to the vector's exclusive size, but would be good to have a test.

ajor (commented):
I'm wondering if this can be made more consistent with the existing code and simplified a little.

Instead of having these `memberSizes` variables at the function-scope and updating `exclusiveSize` at the end, can we just write to `node.exclusiveSize` directly when processing a node, like we do for `staticSize` and `dynamicSize`?

If `processContainer` can write to it directly then that removes the special return value only used for for exclusive size.

jgkamat:
Alastair Robertson writes:

> Instead of having these `memberSizes` variables at the function-scope and 
> updating `exclusiveSize` at the end, can we just write to 
> `node.exclusiveSize` directly when processing a node, like we do for 
> `staticSize` and `dynamicSize`?

The main reason I ended up going with this approach is that we need the 
finalized total size in order to make the exclusive size calculation (as we 
are using exclusiveSize = Total(Node) - Total(sum(members))), and validating 
that might be a bit tricky in all cases. In addition, since we are using 
unsigned variables, doing something like subtracting from a the variable and 
setting it would need a bounds check - and I'd rather have that only in one 
place rather than all over the function (wherever we set exclusiveSize).

Another option that I thought of was summing to exclusiveSize as this patch 
currently does for memberSizes right now, and correcting it at the very end - 
I disliked that because the variable name dosen't match the contents until the 
very end (let me know if you think this is better, though).

I could very well be missing what you're suggesting though - if you can find a 
way to make this simpler let's definitely go for it, because (in my opinion) 
this is a bit of a mess as is :).

ajor:
I was thinking originally that if we put this bounds checking logic could be put into its own function to be used everywhere, e.g.:
```
void calcExclusiveSize(Node &node, uint64_t memberSizes) {
  node.exclusiveSize = ...
}
...
node.dynamicSize = ...
calcExclusiveSize(node, child.staticSize + child.dynamicSize);
```

The early exits from `processContainer()` do make this harder though since there's no single point where all the sizes will be valid - it'd need to be run after each time `node.dynamicSize` is set - doable but could be prone to errors.

One way we could ensure that exclusive size is always set when dynamic size is set would be by only allowing them to be updated through a function that sets both at the same time, e.g.:
```
void setSize(Node &node, uint64_t dynamicSize, uint64_t memberSizes) {
  node.dynamicSize = dynamicSize;
  node.exclusiveSize = ...
}
```

src/TreeBuilder.cpp
```diff
@@ -390,6 +409,9 @@ TreeBuilder::Node TreeBuilder::process(NodeID id, Variable variable) {
           << "', kind: " << drgnKindStr(variable.type) << ")"
           << (variable.isStubbed ? " STUBBED" : "")
           << (th->knownDummyTypeList.contains(variable.type) ? " DUMMY" : "");
+  // Default dynamic size to 0 and calculate fallback exclusive size
+  setSize(node, 0, 0);
+  // Keep track of total sizes of all members
```
ajor, line 414:
```suggestion
```
This comment got left behind from the previous version

src/TreeBuilder.cpp
```diff
@@ -521,6 +548,9 @@ TreeBuilder::Node TreeBuilder::process(NodeID id, Variable variable) {
   return node;
 }
 
+/*
+ * Process a custom container, and return the size of all members.
```
ajor, line 552: This comment is also from the previous version

ajor (approved):
Looks good to me!

src/TreeBuilder.cpp
jgkamat: Ah, I really need to be better at this, I keep forgetting. I'll fix these and land later today :)

PR #100, opened 2023-03-16
Warnings cleanup

JakeHillion:
## Summary

- Disables warnings for test code where they're reasonable.
- Changed something which used bitfields with a default initialiser in our single C++17 file to using whole bools with a default initialiser.

I gave up on turning on `-Werror` because I am weak. Before we turn this on we need to build a consistent set of warnings in all the different build environments, otherwise it's very annoying.

## Test plan

- `make test-devel` - there are fewer warnings.
- CI

### DIFF
--- a/test/integration/arrays.toml
+++ b/test/integration/arrays.toml
@@ -2,9 +2,16 @@ definitions = '''
   struct Foo10 {
     int arr[10];
   };
+
+  #pragma clang diagnostic push
+  #pragma clang diagnostic ignored "-Wzero-length-array"
   struct Foo0 {
     int arr[0];
   };
+
+  using ZeroLengthIntArray = int[0];
+  #pragma clang diagnostic pop
+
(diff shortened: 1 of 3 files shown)


test/integration/arrays.toml
```diff
@@ -72,6 +79,6 @@ definitions = '''
     # WARNING: zero-length arrays are handled differently to non-empty arrays.
     # They end up not being treated as containers. This should probably change
     # in the future.
-    param_types = ["const int(&)[0]"]
+    param_types = ["const ZeroLengthIntArray&"]
```
JakeHillion, line 82:
```
using ZeroLengthIntArray = int[0];
std::cout << std::is_same<const int(&)[0], const ZeroLengthIntArray&>() << std::endl;
```

godbolt prints 1 so pretty sure this is equivalent

ajor (approved):

PR #101, opened 2023-03-16
oil: remove fields in options

JakeHillion:
## Summary

5 of the fields in the OIL options relate to caching and JIT compilation. Remove these as they don't make sense anymore - you should use compile time OIL if you want caching and to disable JIT. These options are not used under compile time OIL.

## Test plan

- `make test-devel`
- CI
- D44216203

### DIFF
--- a/src/OILibraryImpl.cpp
+++ b/src/OILibraryImpl.cpp
@@ -34,19 +34,6 @@ extern "C" {
 
 namespace ObjectIntrospection {
 
-/**
- * We need a way to identify the BLOB/Object file, but we don't have access
- * to the function's name or the introspected type's name without
- * initialising drgn.
- * Exclusively to OIL, we won't use the type's name, but the templated
- * function address. We assume the address of the templated function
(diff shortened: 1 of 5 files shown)


src/OILibraryImpl.cpp
```diff
@@ -152,114 +144,53 @@ int OILibraryImpl::compileCode() {
...
+  }
+  const char *name = drgn_symbol_name(sym);
+  drgn_symbol_destroy(sym);
 
-    auto rootType = symbols->getRootType(irequest{"entry", name, "arg0"});
-    if (!rootType.has_value()) {
-      LOG(ERROR) << "Failed to get type of probe argument";
-      return Response::OIL_COMPILATION_FAILURE;
-    }
+  auto rootType = symbols->getRootType(irequest{"entry", name, "arg0"});
```
JakeHillion, line 162: Add a comment about changing this to the new `drgn_symbol -> drgn_type` pipeline.

ajor (approved):

PR #102, opened 2023-03-17
Migrate `folly` and `rocksdb` to FetchContent

JakeHillion:
## Summary

Move two of our `extern` repos to `FetchContent`, leaving just `drgn`. Keep the existing manual build steps for now. This change is effectively a no-op, completing exactly the same build as before but giving cmake more knowledge of it.

I left `drgn` alone as we change the content a lot, so moving it out of `extern` would slow development.

## Test plan

- `rm -rf build && make test-devel`
- CI


JakeHillion:
Need to create the `testing.oid.toml` dynamically as CMake now decides where folly is.

ajor (approved):

PR #103, opened 2023-03-22
codegen: carry decl and func with containerinfo

### LINKED ISSUE #81: Specifying the same treebuilder config but different code in the `.toml` causes unexpected behaviour
The container `.toml` should carry all but the treebuilder logic for a type. However, due to implementation details, there is currently only one code provided per enum that treebuilder expects. This means that although the container `.toml` can have different code bodies but the same treebuilder, only one of the codes is used. Fix this by carrying the `ContainerInfo` around, perhaps in a shared pointer, rather than using the string map.

JakeHillion:
## Summary

Code generation of container function declarations and definitions currently runs through two maps, `typeToDeclMap` and `typeToFuncMap`. These are indexed by the `ContainerTypeEnum` which comes from a field in `ContainerInfo`. Rather than having this indirection, we should get the `decl` and `func` directly from the `ContainerInfo` object.

- Delete the `ContainerInfo` copy constructor as it now contains potentially very large strings.
- Change all copies of `ContainerInfo` to a reference to `OICodegen`'s `containerInfoList`.
- Remove `ContainerInfo` from `Serialize.h` as we no longer need to serialize it - only the numeric enum passes between `OICodeGen` and `TreeBuilder`, which makes sense!

Specifically, instead of looking up `typeToDeclMap[cinfo.ctype]` we now use `cInfo.codegen.decl` directly. The `.toml` files and code flow haven't really changed, except to avoid copying the types which are now quite big.

Closes #81

## Test plan

- CI
- `make test-devel`

Removed the serialization of `ContainerInfo` from CPP on `main`. Compilation fails with a tonne of static asserts, so I'm confident this is safe.

### DIFF
--- a/src/Common.h
+++ b/src/Common.h
@@ -25,7 +25,54 @@ extern "C" {
 
 constexpr int oidMagicId = 0x01DE8;
 
-struct ContainerInfo;
+#define LIST_OF_CONTAINER_TYPES  \
+  X(UNKNOWN_TYPE)                \
+  X(ARRAY_TYPE)                  \
+  X(SMALL_VEC_TYPE)              \
+  X(SET_TYPE)                    \
+  X(UNORDERED_SET_TYPE)          \
+  X(SEQ_TYPE)                    \
+  X(LIST_TYPE)                   \
+  X(STD_MAP_TYPE)                \
+  X(STD_UNORDERED_MAP_TYPE)      \
+  X(MAP_SEQ_TYPE)                \
+  X(BY_MULTI_QRT_TYPE)           \
+  X(F14_MAP)                     \
+  X(F14_SET)                     \
+  X(FEED_QUICK_HASH_SET)         \
+  X(FEED_QUICK_HASH_MAP)         \
+  X(RADIX_TREE_TYPE)             \
+  X(PAIR_TYPE)                   \
+  X(STRING_TYPE)                 \
+  X(FOLLY_IOBUF_TYPE)            \
+  X(FOLLY_IOBUFQUEUE_TYPE)       \
+  X(FB_STRING_TYPE)              \
+  X(UNIQ_PTR_TYPE)               \
+  X(SHRD_PTR_TYPE)               \
+  X(FB_HASH_MAP_TYPE)            \
+  X(FB_HASH_SET_TYPE)            \
+  X(FOLLY_OPTIONAL_TYPE)         \
+  X(OPTIONAL_TYPE)               \
+  X(TRY_TYPE)                    \
+  X(REF_WRAPPER_TYPE)            \
+  X(SORTED_VEC_SET_TYPE)         \
+  X(REPEATED_FIELD_TYPE)         \
+  X(CAFFE2_BLOB_TYPE)            \
+  X(MULTI_MAP_TYPE)              \
+  X(FOLLY_SMALL_HEAP_VECTOR_MAP) \
+  X(CONTAINER_ADAPTER_TYPE)      \
+  X(MICROLIST_TYPE)              \
+  X(ENUM_MAP_TYPE)               \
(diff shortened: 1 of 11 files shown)


src/ContainerInfo.h
```diff
@@ -73,6 +74,33 @@ ContainerTypeEnum containerTypeEnumFromStr(std::string& str);
 const char* containerTypeEnumToStr(ContainerTypeEnum ty);
 
 struct ContainerInfo {
+  struct Codegen {
+    std::string decl;
+    std::string func;
+  };
+
+  ContainerInfo(const ContainerInfo&) = delete;
```
ajor, line 82: Also delete the copy assignment constructor?

https://en.cppreference.com/w/cpp/language/rule_of_three

src/OICodeGen.h
```diff
@@ -114,8 +114,11 @@ class OICodeGen {
   Config config{};
   FuncGen funcGen;
 
-  using ContainerTypeMap =
-      std::pair<ContainerInfo, std::vector<drgn_qualified_type>>;
+  using RefSharedContainerInfo =
```
ajor, line 117: What does a reference_wrapper of a shared_ptr provide? Just an optimisation to avoid bumping the refcount?

ajor (commented):
shared_ptrs seem unnecessary for us here and the lifetimes of the objects could be made more understandable by changing who owns them.

If these objects are needed by both OICodeGen and TreeBuilder, maybe they should be owned at the level that controls these two pieces (OIDebugger)?

src/ContainerInfo.h
JakeHillion: Always useful having someone who knows what they're doing review C++ :)

I'm a little confused now though as I'm almost certain I don't need a custom destructor, I just want to avoid accidental expensive copies. Is this me misreading "requires" from the docs?

src/OICodeGen.h
JakeHillion: Yep. I was aiming for `&` for just reading, `&shared_ptr` for reading and potentially copying, and `shared_ptr` for owning. In theory this is quicker in a multithreaded context but as we're currently single threaded it's not, so it was a bit of future proofing. In hindsight it was most of the work of this diff to make the really awkward reference-wrappers work...

src/ContainerInfo.h
JakeHillion: Oh, I think this bit explains it:

> Classes that manage non-copyable resources through copyable handles may have to declare copy assignment and copy constructor private and not provide their definitions or define them as deleted. This is another application of the rule of three: deleting one and leaving the other to be implicitly-defined will most likely result in errors.

JakeHillion:
> shared_ptrs seem unnecessary for us here and the lifetimes of the objects could be made more understandable by changing who owns them.
> 
> If these objects are needed by both OICodeGen and TreeBuilder, maybe they should be owned at the level that controls these two pieces (OIDebugger)?

Well now I'm annoyed. It turns out the only thing we do with `ContainerInfo` in `TreeBuilder` is `kind = containerInfo->ctype;`, so it doesn't even need to see the object let alone own it.

I'm thinking of changing everything to `&` using `std::reference_wrapper`, which should be easy.

src/Common.h
```diff
@@ -15,6 +15,7 @@
  */
 #pragma once
 #include <map>
+#include <memory>
```
ajor, line 18:
```suggestion
```
Don't think this is needed in the latest version of this PR

src/ContainerInfo.cpp
```diff
@@ -20,6 +20,8 @@
 
 #include <map>
 
+namespace fs = std::filesystem;
```
ajor, line 23:
```suggestion
```

ajor (approved):

PR #107, opened 2023-03-23
Quit oid when upload/download is requested but cache is disabled

jgkamat:
## Summary
Quit oid if cache is not fully configured. I've tested both sides of this if statement manually internally, and with --mode prod, to make sure other cases are not affected.

### DIFF
--- a/src/OID.cpp
+++ b/src/OID.cpp
@@ -293,6 +293,9 @@ static ExitStatus::ExitStatus runScript(const std::string &fileName,
 
   oid->setCacheRemoteEnabled(oidConfig.cacheRemoteUpload,
                              oidConfig.cacheRemoteDownload);
+  if (!oid->validateCache()) {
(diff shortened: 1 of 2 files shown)


src/OID.cpp
```diff
@@ -293,6 +294,9 @@ static ExitStatus::ExitStatus runScript(const std::string &fileName,
 
   oid->setCacheRemoteEnabled(oidConfig.cacheRemoteUpload,
                              oidConfig.cacheRemoteDownload);
+  if (!oid->validateCache()) {
+    return ExitStatus::CacheConfigurationError;
```
ajor, line 298: Is this not just a "UsageError"?

ajor (approved):

src/OID.cpp
jgkamat: Yup, thanks, I missed that!

PR #108, opened 2023-03-24
std::variant: change to use std::visit

JakeHillion:
## Summary

`getSizeType` for `std::variant` previously used a custom templated visitor. Update to use `std::visit`. This is a `.toml` only change as the data written to the data segment is identical.

`if (container.index() >= sizeof...(Types))` is necessary as `container.valueless_by_exception()` does not work correctly for variants whose index is optimised to a type which can't contain `0xffffffffffffffff`. There is already a test that checks the last possible element is handled correctly.

## Test plan

- `make test-devel`
- CI

### DIFF
--- a/types/std_variant.toml
+++ b/types/std_variant.toml
@@ -11,32 +11,24 @@ void getSizeType(const %1%<Types...> &container, size_t& returnArg);
 """
 
 func = """
-template <size_t I = 0, class... Types>
-void getSizeVariantContents(const %1%<Types...> &container, size_t& returnArg)
-{
-  if constexpr (I < sizeof...(Types))
-  {
-    if (I == container.index())
-    {
-      // Contents are stored inline - don't double count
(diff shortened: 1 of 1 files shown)


ajor (approved):
Much nicer, thank you!

Would you be able to add a comment about the valueless_by_exception / integer width problem please (like I should have done months ago)? Maybe it'll help the next poor soul who has to deal with std::variant

JakeHillion:
> Much nicer, thank you!
> 
> Would you be able to add a comment about the valueless_by_exception / integer width problem please (like I should have done months ago)? Maybe it'll help the next poor soul who has to deal with std::variant

The weird thing is I couldn't reproduce this with quite a few versions on `godbolt`, `valueless_by_exception` returned correctly for any size. Either way I'll document it!

ajor:
One theory I had but didn't follow up on was that we could be using a different standard library in OI than we use to compile the test programs, i.e. libc++ vs stdlibc++

PR #109, opened 2023-03-24
Avoid following weak_ptrs

jgkamat:
## Summary

Previously, we treated weak_ptrs as normal types and we recursed within them, following the internal data pointer and possibly causing crashes. We really shouldn't be following them, so I added a custom type to simply abort processing. If we want to handle them (ie: check if they are valid, and follow them if so), that should be fairly easy with the work there is here so far.

If people think we should be following weak pointers, I'm happy to add that in.

## Test plan
Added test cases, and manual testing on a couple types (verifying the json output).


ajor (approved):
Looks good!

JakeHillion (approved):
great stuff :)

PR #110, opened 2023-03-27
tests: build integration.py targets with cmake

JakeHillion:
## Summary

After removing `folly` from `extern` the `integration.py` targets can no longer build. Move them into `test/CMakeLists.txt` to make the `folly` dependency explicit.

- Build the two integration targets with CMake with an explicit `folly_headers` dependency.
- Bring `integration_py` into `ctest`, gated with `WITH_FLAKY_TESTS`.
- Skip `test_data_segment_size_change` which was consistently failing on my machine and I understand is broken.

## Test plan

- `make test-devel`
- CI


CMakeLists.txt
```diff
@@ -123,6 +123,7 @@ FetchContent_Populate(folly)
 
 add_library(folly_headers INTERFACE)
 target_include_directories(folly_headers SYSTEM INTERFACE ${folly_SOURCE_DIR})
+target_link_libraries(folly_headers INTERFACE Boost::headers)
```
ajor, line 126: Do you know why this is needed now? There don't seem to be any references to boost in the old Makefile

ajor (changes requested):

CMakeLists.txt
JakeHillion: I think CMake is doing the dependency management properly here. Boost is in `/usr/include/boost/` for me so a C++ compiler would be able to find it without any extra paths, but CMake seems to want it explicitly specified. It definitely fails to build with just Folly headers as `folly::Functional` depends on `expr_iif` or something in Boost.

test/integration.py
```diff
@@ -41,24 +41,24 @@ def setUp(self):
...
-        self.temp = tempfile.TemporaryDirectory(
-            dir=os.path.join(self.oid_source_dir, "build/")
-        )
+        self.temp = tempfile.TemporaryDirectory(dir=self.build_dir)
         os.chdir(self.temp.name)
 
-        self.oid = os.path.join(self.oid_source_dir, "build/oid")
-        self.oid_conf = os.path.join(self.oid_source_dir, "build/testing.oid.toml")
+        self.oid = os.getenv("OID")
+        self.oid_conf = os.path.join(self.build_dir, "..", "testing.oid.toml")
```
JakeHillion, line 55: I left this one as is because it was already doing it and this file is only put there by the `Makefile` anyway. We should probably fix that but I see it as a separate problem.

ajor (approved):

PR #116, opened 2023-03-31
Display the logo

samuelnair:
Would be good to display the logo on the landing page.

## Summary
Describe what your change accomplishes

## Test plan
Check that OI is working correctly, you can add some unit or integration test,
paste the output of some manual test, and / or paste the output of running
the test locally with `make test-static`


tyroguru (approved):
Good idea! Thanks for doing it!

PR #118, opened 2023-04-03
Build rocksdb within CMake rather than with a subcommand

JakeHillion:
## Summary

Build rocksdb within CMake. This means that we only have to build rocksdb once on a clean build, speeding up repeated iterations of e.g. `make oid-devel`, and increasing the speed of test driven development.

## Test plan

- it's not


JakeHillion:
realates to #131

Issue #95, opened 2023-03-15
Typed TreeBuilder


JakeHillion:
# Typed TreeBuilder

This piece of work aims to solve four problems, in this priority order:
1. Level up the output from Object Introspection as a Library (OIL) to the same level as OI Debugger, enabling new use cases.
2. Unify the code path that OIL and OID take in our generated code, increasing maintainability.
3. Prevent desynchronisation between the logic that writes to the data segment (generated code) and the logic that reads from it (TreeBuilder/OIL consumers), similarly increasing maintainability.

While these may be a lofty set of goals, I believe by taking this approach we can turn them into one cohesive piece of work.

## Glossary

| **Term** | **Definition** |
| -------- | -------------- |
| OI | Object Introspection. A system of debug analysis and code generation that enables traversal and information gathering from arbitrary C++ types. |
| OID | Object Introspection Debugger. The current main deployment of OI, attached externally to a running process as a debugger. Produces a rich output detailing the captured object in RocksDB or JSON format. |
| OIL | Object Introspection as a Library. A C++ interface to the OI technology, available in a JIT mode or a compile time mode. Produces a limited output describing the size of the object. |
| Statically Typed | A type of an object known entirely at compile time, and erased before runtime. |
| Dynamically Typed | A type of an object unknown at compile time, and instead explored at runtime. |

## Motivation

OI's traversal of arbitrary type hierarchies presents many opportunities, including: opportunities for compression, computing object differentials at a primitive level, and analysis based on captured memory.

OIL in its current form uses an accumulator to track each element which contributes to the total size. OID instead utilises a data segment to send information about the captured object from the target to the attached debugger. It is very easy to have OIL return instead a `std::vector<uint8_t>` which contains the contents of this data segment, providing the ability to access this already rich data in the end-user application. This is feasible because OIL runs on request in the target process, and unlike OID can utilise the memory allocator for a dynamic array.

While an array of bytes is a great way to store the data efficiently and share between processes, it provides an incredibly poor interface for the consumer. Our current consumer is [TreeBuilder](https://github.com/facebookexperimental/object-introspection/blob/de071e2e38420c3cb2a0931c418be8044ee16255/src/TreeBuilder.cpp), a behemoth which infers the contents of the data segment in an entirely different way to the producer. That is, the generated code writes to the data segment based on dynamic logic for containers in `.toml` files, while `TreeBuilder` reads from the data segment using a statically compiled set of logic.

[...]

PR #121, opened 2023-04-05
Cycles

### LINKED ISSUE #104: Cycle detection broken in OIL
Our pointer cycle detection currently relies on both JIT code and TreeBuilder to fully stop processing cycles.

This works in OID, but means that OIL ends up processing some nodes in a cycle (just the first?) more than once as it lacks the TreeBuilder component.

This seems like it should be able to work fully for both OID and OIL without the TreeBuilder component.

See test/integration/cycles.toml for tests which fail under OIL.

### Code References
Core JIT cycle detection code: https://github.com/facebookexperimental/object-introspection/blob/c4194a9fbc92f7ed0f14aec2d81e26b8f0c411fb/src/OITraceCode.cpp#L35-L74
Unique pointer cycle detection, making use of the above: https://github.com/facebookexperimental/object-introspection/blob/c4194a9fbc92f7ed0f14aec2d81e26b8f0c411fb/types/uniq_ptr_type.toml#L26

TreeBuilder has its own independent cycle detection logic: https://github.com/facebookexperimental/object-introspection/blob/c4194a9fbc92f7ed0f14aec2d81e26b8f0c411fb/src/TreeBuilder.cpp#L370-L377

JakeHillion:
## Summary

Moves all of the pointer detection logic out of TreeBuilder and into the data segment. Closes #104 as there is now no opportunity for OIL to have a different issue in pointer detection logic to OID. This makes it easier to type the datasegment as part of #95. There is room for improvement as null pointer implies the following will be a zero, meaning we could save a byte, but this is easier to implement for an MVP of the typed data segment.

Also fixes a bug in the `integration.py` tests where the config wasn't passed to OID, causing it to use the default. This took a lot of head scratching to debug.

## Test plan

- `make test-devel`
- CI
- Tried adding a test for `folly::IOBufQueue`. Unfortunately it's completely broken, because of #123 and not including the logic for traversing the `folly::IOBuf`, at minimum.

### DIFF
--- a/src/TreeBuilder.cpp
+++ b/src/TreeBuilder.cpp
@@ -40,7 +40,7 @@ extern "C" {
 }
 
 /* Tag indicating if the pointer has  been followed or skipped */
-enum class TrackPointerTag: uint64_t {
+enum class TrackPointerTag : uint64_t {
   /* The content has been skipped.
    * It prevents double counting the footprint of a node the JIT code has seen
    * before, double counting content being stored inline, or getting stuck in an
@@ -225,9 +225,6 @@ void TreeBuilder::build(const std::vector<uint64_t>& data,
     auto& rootID = rootIDs.emplace_back(nextNodeID++);
 
     try {
-      // The first value is the address of the root object. Drop it as we don't
-      // need to manage pointer deduplication anymore.
-      next();
       process(rootID, {.type = type, .name = argName, .typePath = argName});
     } catch (...) {
       // Mark the failure using the error node ID
--- a/src/TreeBuilder.cpp
+++ b/src/TreeBuilder.cpp
@@ -39,6 +39,17 @@ extern "C" {
 #include <sys/types.h>
 }
 
+/* Tag indicating if the pointer has  been followed or skipped */
+enum class TrackPointerTag: uint64_t {
+  /* The content has been skipped.
+   * It prevents double counting the footprint of a node the JIT code has seen
+   * before, double counting content being stored inline, or getting stuck in an
+   * infinite loop when processing circular linked list.
+   */
+  skipped = 0,
+  followed = 1,
+};
+
 TreeBuilder::TreeBuilder(Config c) : config{std::move(c)} {
   buffer = std::make_unique<msgpack::sbuffer>();
 
@@ -205,7 +216,6 @@ void TreeBuilder::build(const std::vector<uint64_t>& data,
   th = &typeHierarchy;
   oidData = &data;
 
-  pointers.clear();
   oidDataIndex = 3;  // HACK: OID's first 3 outputs are dummy 0s
 
   ObjectIntrospection::Metrics::Tracing _("build_tree");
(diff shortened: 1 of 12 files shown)


src/TreeBuilder.cpp
```diff
@@ -215,8 +214,9 @@ void TreeBuilder::build(const std::vector<uint64_t>& data,
     auto& rootID = rootIDs.emplace_back(nextNodeID++);
 
     try {
-      // The first value is the address of the root object
-      pointers.insert(next());
+      // The first value is the address of the root object. Drop it as we don't
+      // need to manage pointer deduplication anymore.
+      next();
```
ajor, line 219: Can we just stop storing this pointer in the first place?

src/TreeBuilder.cpp
```diff
@@ -741,7 +731,7 @@ void TreeBuilder::processContainer(const Variable& variable, Node& node) {
         // Contents aren't actually stored inline in this case,
         // but we set this to `true` so that we don't double-count
         // this string's data if we have seen it before.
-        contentsStoredInline = !shouldProcess(*node.pointer);
+        contentsStoredInline = (*node.pointer != 0);
```
ajor, line 734: How does this work? Won't node.pointer always be set?

ajor (commented):
I think it might make the code clearer about what it's doing if we wrap the `next() == 0` checks away in a function with a descriptive name, e.g. `bool shouldFollowPointer() { return next() == 1; }`

src/TreeBuilder.cpp
JakeHillion: It's completely wrong - my bad! It also fixes that test I skipped once this is done properly.

src/TreeBuilder.cpp
JakeHillion: Yes. I was a little worried as I'm not sure I've got every entrypoint where it's written, but I put it in a separate commit so it can be easily reverted.

JakeHillion:
> I think it might make the code clearer about what it's doing if we wrap the `next() == 0` checks away in a function with a descriptive name, e.g. `bool shouldFollowPointer() { return next() == 1; }`

I went for `if (bool followPointer = next() == 1; !followPointer)` as I think it solves the clarity issue without obfuscating the `next()` call (imo that makes it much harder to debug desyncs). In new treebuilder it'll be a `Sum` anyway which should clear things up.

src/TreeBuilder.cpp
```diff
@@ -626,15 +616,15 @@ void TreeBuilder::processContainer(const Variable& variable, Node& node) {
       node.pointer = next();
       containerStats.length = *node.pointer ? 1 : 0;
       containerStats.capacity = 1;
-      if (!shouldProcess(*node.pointer)) {
+      if (bool followPointer = next() == 1; !followPointer) {
```
ttreyer, line 619: I have a hard time _following_ this line... The negation makes it confusing to understand at a glance.
Using some constants could help in this case. What about something like:
```cpp

/* Tag indicating if the pointer has  been followed or skipped */
enum class TrackPointerTag: uintptr_t {
  /* The content has been skipped.
   * It prevents double counting the footprint of a node the JIT code has seen before,
   * double counting content being stored inline, or getting stuck in an infinite loop
   * when processing circular linked list.
   */
  skipped = 0,
  followed = 1,
};

if (next() == TrackPointerTag::skipped) return;
```

PR #124, opened 2023-04-18
Generate a shared library linked to integration_test_target

ttreyer:
## Summary
Extend our integration test suite by linking a generated shared library to the integration_test_target. This enable us testing and validating that `oid` handles shared libraries. We can add definitions in the shared library using the keys `definitions_shlib` and `raw_definitions_shlib`.

## Test plan
There are no tests using the new shlib definitions, yet. However, we can see this PR is not affecting our existing tests:
```
$ make configure-devel
[...]

$ make test
[...]
100% tests passed, 0 tests failed out of 420

Total Test time (real) =  75.04 sec
```

### DIFF
--- a/test/integration/gen_tests.py
+++ b/test/integration/gen_tests.py
@@ -250,6 +250,32 @@ def gen_target(output_target_name, test_configs):
         add_footer(f)
 
 
+def gen_target_shlib(output_target_name, test_configs):
+    with open(output_target_name, "w") as f:
+        headers = set()
+        for config in test_configs:
+            headers.update(config.get("includes_shlib", []))
+        add_headers(f, custom_headers=headers, thrift_headers=[])
+
+        from textwrap import dedent
+
+        for config in test_configs:
+            ns = config["suite"]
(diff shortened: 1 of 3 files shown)


test/integration/gen_tests.py
```diff
@@ -250,6 +250,32 @@ def gen_target(output_target_name, test_configs):
         add_footer(f)
 
 
+def gen_target_shlib(output_target_name, test_configs):
+    with open(output_target_name, "w") as f:
+        headers = set()
+        for config in test_configs:
+            headers.update(config.get("includes_shlib", []))
```
ajor, line 257: Shall we just use the same "includes" for the shared lib? It'd simplify writing the test configs and the worst that'll happen is that we'll get unnecessary includes

ajor (approved):

test/integration/gen_tests.py
ajor: Otherwise, "includes_shlib" should be documented too

PR #125, opened 2023-04-18
Add a feature system to OI to make it easier to add optional features in the future

JakeHillion:
## Summary

Adds features which are specified either on the command line or the config.

- Enable a feature with e.g. `-fchase-raw-pointers`.
- Disable a feature with e.g. `-Fchase-raw-pointers`, last on the command line is always preferred.
- Add a base set of features in the config using: `features = ["chase-raw-pointers"]`.
- Messed around with the test extra config option as sometimes with `.toml` things need to be at the top or at the bottom of the file.

## Test plan

- New tests for each case.
- Checked `oilgen` still functions on its limited test case (really should write proper external tests for that...).
- `make test`
- CI

### DIFF
--- a/src/OID.cpp
+++ b/src/OID.cpp
@@ -22,15 +22,18 @@
 #include <cstdlib>
 #include <filesystem>
 #include <iostream>
+#include <map>
 
 extern "C" {
 #include <getopt.h>
 #include <libgen.h>
 }
 
+#include "Features.h"
 #include "Metrics.h"
 #include "OIDebugger.h"
 #include "OIOpts.h"
+#include "OIUtils.h"
 #include "PaddingHunter.h"
 #include "TimeUtils.h"
 #include "TreeBuilder.h"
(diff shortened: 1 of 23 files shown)


src/OID.cpp
```diff
@@ -463,12 +478,13 @@ int main(int argc, char* argv[]) {
   std::string configGenOption;
   std::optional<fs::path> jsonPath{std::nullopt};
 
+  std::map<Feature, bool> features = {
+      {Feature::PackStructs, true},
```
JakeHillion, line 482: I didn't really get to the bottom of what this does (and why we'd ever want to disable it), but I left it there for now for consistency.

tyroguru:
General comment: I really like this! Quite happy with the -f/-F enable/disable feature option. Why are we keeping the exisiting command line options though, e.g., -P? Isn;t one of the beauties of this that we get to cleanup the option space?

JakeHillion:
> General comment: I really like this! Quite happy with the -f/-F enable/disable feature option. Why are we keeping the exisiting command line options though, e.g., -P? Isn;t one of the beauties of this that we get to cleanup the option space?

Yes, it definitely is. I left them there for now for backwards compatibility: I don't particularly fancy tracking down every call of OID to clean them up. They can definitely go in the future.

tyroguru (approved):
I really like this as I have previously said. As you have previously said Jake, we need to remove the other options in conjunction with Strobelight changes.

PR #126, opened 2023-04-18
clang-format: set BinPackParameters=false

JakeHillion:
## Summary

When most arguments in a function definition are on separate lines but some are sneaked onto the end I miss them, and it's caused me to get quite confused several times. This change means that if the list of parameters don't fit on one line then each parameter will have its own line. It also makes diffs easier to parse.

Anyone, feel free to veto this :) - I'll leave it up for a little while.

## Test plan

- no-op


tyroguru (approved):
After discussion yesterday I think we should land this and see how we get on with it. The reasons Jake provides for it make sense so let's see how we live with it.

PR #127, opened 2023-04-19
Faster locateSymbol by re-using dwfl across calls

ttreyer:
## Summary
SymbolService is used to locate symbols when the JIT code is being relocated in the target process. Our previous implementation was inefficient, as a new `Dwfl` instance was created, initialised with the target's modules, then destroyed, for each call to `locateSymbol`.

This PR modifies SymbolService to keep the same `Dwfl` instance across multiple calls to `locateSymbol`. Dwfl is initialised once, during the SymbolService construction.

## Test plan
This PR doesn't change `oid`'s behaviour. The tests should be enough to validate the changes.
```
$ make test
[...]
100% tests passed, 0 tests failed out of 420

Total Test time (real) =  66.51 sec
```

### DIFF
--- a/src/Common.h
+++ b/src/Common.h
@@ -105,3 +105,14 @@ struct TypeHierarchy {
   std::set<struct drgn_type*> thriftIssetStructTypes;
   std::map<struct drgn_type*, std::vector<struct drgn_type*>> descendantClasses;
 };
+
+// Helper for std::variant and std::visit
+// https://en.cppreference.com/w/cpp/utility/variant/visit
+template <class... Ts>
+struct visitor : Ts... {
+  using Ts::operator()...;
+};
+
+// Type deduction for the helper above
+template <class... Ts>
+visitor(Ts...) -> visitor<Ts...>;
--- a/src/SymbolService.cpp
+++ b/src/SymbolService.cpp
@@ -85,16 +85,28 @@ static bool isExecutableAddr(
   return it != end(exeAddrs) && addr >= it->first;
 }
 
-SymbolService::SymbolService(std::variant<pid_t, fs::path> newTarget) {
-  target = std::move(newTarget);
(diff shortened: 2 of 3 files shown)


src/SymbolService.cpp
```diff
@@ -213,55 +218,70 @@ std::optional<SymbolInfo> SymbolService::locateSymbol(
...
-
-      // Add module's boundary to executableAddrs
-      executableAddrs = {{start, end}};
-
-      break;
-    }
+  dwfl_report_begin(dwfl);
+
+  bool ok = std::visit(
+      visitor{[this](pid_t target) -> bool {
```
JakeHillion, line 230: Could we make these two methods with descriptive names? I initially had quite a hard time following this. E.g. `loadModulesFromPid` and `loadModulesFromExecutable`.

ttreyer: I wish I didn't have to use `std::visit`, but it's the only type-safe way of ensuring all _variants_ are covered 😞

tyroguru:
Thanks Thierry for the explanation of how this works and I'm happy with that. However, *never* remove the comment in the code which includes a link to a page which details how the visitor works. This is seriously non-intuitive.

tyroguru (approved):
Just left the one comment. Happy for it to merge.

src/Common.h
```diff
@@ -105,3 +105,14 @@ struct TypeHierarchy {
...
+
+// Helper for std::variant and std::visit
+// https://en.cppreference.com/w/cpp/utility/variant/visit
+template <class... Ts>
+struct visitor : Ts... {
+  using Ts::operator()...;
+};
+
+// Type deduction for the helper above
+template <class... Ts>
```
JakeHillion, line 117: nit: is this necessary? The only file this is used in should be C++20, and the cppreference page says the explicit type deduction is not needed under C++20.

PR #130, opened 2023-04-21
Pass in CodeCov.io token through CI env var

ajor:
CodeCov's docs say that a token isn't required for a public repository, but our uploads have been broken for a few weeks at this point and passing in the token fixes them...


JakeHillion (approved):

ajor:
Environment variables are not being shared with forks, so codecov still fails on this PR. I'm hoping that this PR will at least fix codecov on the main branch though.

PR #133, opened 2023-04-24
features: switch to bitset

JakeHillion:
## Summary

Features were implemented as a `std::set` and cached as bits. Make this more efficient by storing them in a `bitset` (wrapped to have enum accessor).

## Test plan

- `make test`
- CI


ajor (approved):
Looks good!

PR #137, opened 2023-05-03
drgn: refactor, cull dead code, and disable inline

JakeHillion:
## Summary

This is a brutal change to the drgn fork, getting it ready for a rebase. Mainly:
- Removes commits of already dead code.
- Squashes fixes to earlier commits into the earlier commit, hopefully making each commit valid in isolation.
- Disables function inlining and the drgn infrastructure. It's not widely used, and we can cherry-pick it back if it becomes of interest in the future.

All in all this gets us down to 26 commits ahead, and the majority are neat and self contained. 2 are already upstreamed and 1 will be worked on soon.

## Test plan

- `make test`
- CI

### DIFF
--- a/oi/SymbolService.cpp
+++ b/oi/SymbolService.cpp
@@ -500,6 +500,7 @@ static void parseFormalParam(Dwarf_Die& param,
   VLOG(1) << "Adding function arg address: " << farg;
 }
 
+/*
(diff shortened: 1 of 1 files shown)


tyroguru (approved):
Great stuff Jake.

PR #139, opened 2023-05-18
resources: manage headers properly

JakeHillion:
## Summary

Previously we had an `R"(` string in `OITraceCode.cpp` which allowed us to include the file as a string. Instead, keep `OITraceCode.cpp` a fully formed C++ file and utilise the build system to turn it into a string. This will be used for more header files that are needed both as valid headers and as strings for JIT compilation in the Typed TreeBuilder work.

## Test plan

- `make test`
- CI


ajor (approved):

ajor:
CodeCov is back 🎉

PR #142, opened 2023-05-22
Fix all warnings and enable -Werror for all targets

ajor:
The previous method of enabling warnings on a target-by-target basis made it very easy to accidentally miss a target. This had happened in a number of instances, e.g. "codegen" and "symbol_service".

We now instead enable warnings globally, then disable them for a few targets which are for third-party code.


JakeHillion (approved):
nice!

ajor:
Many commits later, I think I've got something that works in the CI 🎉

Issue #143, opened 2023-05-24
Container Alignment is Broken


ajor:
See tests at https://github.com/facebookexperimental/object-introspection/blob/main/test/integration/alignment.toml

We normally deal with alignment by adding explicit padding members to structs. However, this does not work for containers, where we do not control padding ourselves.

We need a way of determining what the alignment of a given type is, and then we must apply an `alignas` specifier to our generated types.

PR #144, opened 2023-05-24
Add Alignment Tests

ajor:
More prep-work for TypeGraph, but useful in general and found a new bug: #143.

Also update the test framework so that OIL tests assume a dynamicSize of 0 when none is explicitly set. This just makes it a little easier to write tests for static types.


JakeHillion (approved):
🧪

PR #145, opened 2023-05-24
OID: Set glog VLOG level globally for all modules

ajor:
It was previously easy to miss modules and I haven't been able to notice any difference in behaviour when setting the VLOG level globally (despite what the old comment says).


JakeHillion (approved):
🪵

PR #148, opened 2023-05-25
TypeGraph: Update container TOML format

ajor:
This change is fully backwards compatible with the old ContainerInfo parser, as it is just adding new fields.
    
The old fields will continue to be used by legacy OICodeGen until that is removed.
    
Also add a README describing the format.

### DIFF
--- a/oi/ContainerInfo.cpp
+++ b/oi/ContainerInfo.cpp
@@ -48,7 +48,7 @@ const char* containerTypeEnumToStr(ContainerTypeEnum ty) {
   }
 }
 
-std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
+[[deprecated]] std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
     const fs::path& path) {
   toml::table container;
   try {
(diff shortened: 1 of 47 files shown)


oi/ContainerInfo.cpp
```diff
@@ -164,10 +165,84 @@ std::unique_ptr<ContainerInfo> ContainerInfo::loadFromFile(
...
-
+      {},
       {
           std::move(decl),
           std::move(func),
       },
   });
 }
+
+ContainerInfo::ContainerInfo(const fs::path& path) {
```
JakeHillion, line 176: I'm personally not a fan of this constructor, but that could be a holdover from other languages. Particularly the:
`const auto &info = containerInfos_.emplace_back(path);` which comes later is very unnatural to me.

Does this approach have significant benefits over changing the signature of the previous to `ContainerInfo ContainerInfo::loadFromFile(...)`? That would also give us the opportunity to use the more explicit `std::expected<ContainerInfo>` in C++23.

It's not a hill I want to die on, but a non-explicit and exception throwing cast from `fs::path&` to `ContainerInfo` feels strange to me.

ajor: There are two separate things at play here:
1. Initialising ContainerInfo in the constructor

    This is idiomatic C++. After calling the ctor it is expected that an object will be fully usable. I would say the the invariant of a ContainerInfo class should be that it represents a container, i.e. has a valid typeName, ctype, etc. There are a few core guidelines I found on this: [C.40](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-ctor) [C.41](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-complete) [C.42](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-throw) [C.50](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-factory).

    This method is also more flexible and more efficient, as it means that ContainerInfo objects can be constructed anywhere, rather than forcing users into having them inside unique_ptrs.

1. Using std::vector's `emplace_back` method

    The new constructor is marked "explicit", so it won't allow implicit casts from a file path to ContainerInfo. The apparent implicitness comes from using `emplace_back`, which forwards the argument from this function call onto ContainerInfo's ctor.

    `v.emplace_back(path)` is similar to writing `v.push_back(ContainerInfo(path))`, except it guarantees that the object will be constructed in place, so is more efficient.

types/README.md
```diff
@@ -0,0 +1,63 @@
...
+
+  C++ code for the declaration of a `getSizeType` function for this container.
+
+- `func`
+
+  C++ code for the definition of a `getSizeType` function for this container.
+
+
+## Changes introduced with TypeGraph
+- `typeName` and `matcher` fields have been merged into the single field `type_name`.
```
JakeHillion, line 40: I don't agree with this change. The idea of `matcher` is to be able to combine `types/string_type.toml` and `types/cxx11_string_type.toml` which are identical excluding the type name, reducing duplication. It hasn't been done yet, but it could get rid of two duplicate type files if used fully. Was there another motivation for removing it?

oi/ContainerInfo.h
```diff
@@ -78,6 +85,10 @@ struct ContainerInfo {
   bool operator<(const ContainerInfo& rhs) const {
     return (typeName < rhs.typeName);
   }
+
+  std::regex getMatcher() const {
+    return std::regex("^" + typeName + "<|^" + typeName + "$");
```
JakeHillion, line 90: It probably makes more sense to keep this as a private field and return a reference, as I understand it this constructor parses the regex so is probably quite expensive.

types/README.md
ajor: My feeling is that `matcher` is too low level to require users to define themselves. Each regex engine has a slightly different syntax, which I think should be an implementation detail hidden from users.

Regex has some potential pitfalls for users too, if they're not careful (and not everyone's a regex expert), e.g. some mistakes I can see happening:
`matcher = "ns::vector<"` matches "other_ns::vector"
`matcher = "^ns::vector\b"` matches "ns::vector::value_type"

For all the types we currently have, it's possible to automatically generate a regex from the type name, which (I hope) will work correctly.

Deduplicating TOML files is a good goal, but again I think there could be some struggles for users to get the right regex syntax. How about making type_name into a list and automatically generating the regex from the multiple names? e.g.:
`type_names = ["std::basic_string", "std::__cxx11::basic_string"]`

ajor:
I moved the creation of the matcher regex back into ContainerInfo's constructor and added some unit tests.

Talking about regexes being hard to write, I found a case my previous one was failing at 😛 . Hopefully that's it working properly now, at least for well-formed types.

PR #151, opened 2023-05-26
TypeGraph: Add new CodeGen

ajor:
## Summary
This code mostly works, but is obviously not complete. This commit just adds the code and tests, but does not enable it in OID or OIL.

## Test plan
All integration tests pass when running against legacy OICodeGen
82% of OidIntegration tests pass when running against TypeGraph CodeGen

### DIFF
--- a/oi/OIDebugger.cpp
+++ b/oi/OIDebugger.cpp
@@ -44,6 +44,7 @@ extern "C" {
 
 #include <glog/logging.h>
 
+#include "oi/CodeGen.h"
 #include "oi/ContainerInfo.h"
 #include "oi/Headers.h"
 #include "oi/Metrics.h"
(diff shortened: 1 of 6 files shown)


oi/OIDebugger.cpp
```diff
@@ -2918,15 +2922,24 @@ std::optional<std::string> OIDebugger::generateCode(const irequest& req) {
...
-  if (auto sourcePath = cache.getPath(req, OICache::Entity::Source)) {
-    std::ofstream(*sourcePath) << code;
-  }
-
   typeInfos.emplace(
       req,
       std::make_tuple(RootInfo{rootInfo.varName, codegen->getRootType()},
                       codegen->getTypeHierarchy(), codegen->getPaddingInfo()));
 
+  if (useTypeGraph) {
```
JakeHillion, line 2930: As this is only in one place, it would probably be cleaner to leave the signature alone and check `generatorConfig.features[Feature::TypeGraph]` explicitly.

JakeHillion (approved):

PR #153, opened 2023-05-30
TypeGraph: Fix multi dimensional arrays

ajor:
Multi dimensional arrays are not flattened into 1-D arrays when using TypeGraph. Update TreeBuilder to account for this.

By not flattening arrays, we are able to produce more descriptive results.

The disadvantage is that we must now recurse inside arrays containing only primitives. A better solution to requiring flattening would be the planned work to not recurse into any static types (not just primitives). This would also apply to multi-dimensional arrays of primtivies.

## Test plan
The following test now passes with and without TypeGraph:
```
OidIntegration.arrays_multidim
```

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -81,6 +81,9 @@ struct OIArray {
 }
 
 void addIncludes(const TypeGraph& typeGraph, std::string& code) {
+  // Required for the offsetof() macro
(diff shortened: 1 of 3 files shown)


JakeHillion (approved):
nice! I like overriding the feature set for these feature specific tests.

PR #155, opened 2023-05-31
Turn FeatureSet into a generic EnumBitset

ajor:
To allow it to be reused with other enums


JakeHillion (approved):
It's really surprising to me that neither of `boost` or `folly` contain this sort of `EnumSet`. Nice!

PR #157, opened 2023-05-31
Typed Data Segment

JakeHillion:
## Summary

This change adds a type system to use when writing to the data segment. Information is serialized into the data segment in the running JIT code, and taken out in TreeBuilder. Creating a type system for this is the first step of ensuring these always align.

Every change here should be gated behind `Feature::TypedDataSegment`, meaning this change has no effect unless enabled.

Major holes:
- Cyclical data types currently do not work as the type can't have loops, for example a linked list that contains a pointer which is either null or points to the tail of the list (a linked list).

Relates to #95 

## Test plan

as of `d198b46`:
- standard: 100% tests passed, 0 tests failed out of 578
- `-ftype-graph`: 94% tests passed, 33 tests failed out of 578
- `-ftyped-data-segment`: 86% tests passed, 80 tests failed out of 578

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -19,6 +19,8 @@
 
 #include <boost/format.hpp>
 #include <iostream>
+#include <set>
+#include <string_view>
 
 #include "oi/FuncGen.h"
 #include "oi/Headers.h"
@@ -51,7 +53,26 @@ void defineMacros(std::string& code) {
     code += R"(
 #define SAVE_SIZE(val)
 #define SAVE_DATA(val)    StoreData(val, returnArg)
+)";
+  } else {
+    code += R"(
+#define SAVE_SIZE(val)    AddData(val, returnArg)
+#define SAVE_DATA(val)
+)";
+  }
+}
 
+void defineArray(std::string& code) {
+  code += R"(
+template<typename T, int N>
+struct OIArray {
+  T vals[N];
+};
+)";
+}
+
+void defineJitLog(std::string& code) {
+  code += R"(
 #define JLOG(str)                           \
   do {                                      \
     if (__builtin_expect(logFile, 0)) {     \
@@ -65,31 +86,25 @@ void defineMacros(std::string& code) {
       __jlogptr((uintptr_t)ptr);        \
     }                                   \
   } while (false)
-
-template<typename T, int N>
-struct OIArray {
-  T vals[N];
-};
 )";
-  } else {
-    code += R"(
-#define SAVE_SIZE(val)    AddData(val, returnArg)
-#define SAVE_DATA(val)
-#define JLOG(str)
-#define JLOGPTR(ptr)
-)";
-  }
 }
 
-void addIncludes(const TypeGraph& typeGraph, std::string& code) {
-  // Required for the offsetof() macro
-  code += "#include <cstddef>\n";
-
-  // TODO deduplicate containers
+void addIncludes(const TypeGraph& typeGraph,
+                 FeatureSet features,
+                 std::string& code) {
+  std::set<std::string_view> includes{"cstddef"};
+  if (features[Feature::TypedDataSegment]) {
+    includes.emplace("functional");
+  }
   for (const Type& t : typeGraph.finalTypes) {
     if (const auto* c = dynamic_cast<const Container*>(&t)) {
-      code += "#include <" + c->containerInfo_.header + ">\n";
+      includes.emplace(c->containerInfo_.header);
     }
+  }
+  for (const auto& include : includes) {
+    code += "#include <";
+    code += include;
+    code += ">\n";
   }
 }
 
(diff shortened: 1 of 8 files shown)


oi/FuncGen.cpp
```diff
@@ -391,3 +430,217 @@ void FuncGen::DeclareGetContainer(std::string& testCode) {
...
+    class DataBuffer {
+      protected:
+        void write_byte(uint8_t);
+    };
+    class DataSegment: public DataBuffer {
+      public:
+        DataSegment(size_t offset) : buf(dataBase + offset) {}
+        void write_byte(uint8_t byte) {
+          if (buf > (dataBase + dataSize * 8)) {
+            // TODO: fail
```
JakeHillion, line 447: TODO: check how the existing method fails and replicate that

ajor:
Don't forget documentation! The rest of the options in the container TOML files are listed in [the README](https://github.com/facebookexperimental/object-introspection/blob/main/types/README.md).

Besides anything else, it'll help me understand what I'm looking at 😅

oi/CodeGen.cpp
```diff
@@ -398,6 +412,140 @@ void addGetSizeFuncDefs(const TypeGraph& typeGraph,
...
+    StaticTypes::Unit<DB>
+    getSizeType(const T &t, typename TypeHandler<DB, T>::type returnArg) {
+      JLOG("obj @");
+      JLOGPTR(&t);
+      return TypeHandler<DB, T>::getSizeType(t, returnArg);
+    }
+  )";
+
+  code += R"(
+    template<typename DB, typename T0, long unsigned int N>
```
ajor, line 428: Why "T0" instead of "T"?

oi/OITraceCode.cpp
```diff
@@ -20,7 +20,7 @@
 #define C10_USING_CUSTOM_GENERATED_MACROS
 
 // These globals are set by oid, see end of OIDebugger::compileCode()
-extern uintptr_t dataBase;
+extern uint8_t* dataBase;
```
ajor, line 23: The types of this variable seem all kinds of crazy in FuncGen...
Does changing its type here help with something?

ajor (commented):
I'm going to have to brush up on what's going on in this code - are the details in #95 still up-to-date?

oi/CodeGen.cpp
JakeHillion: The idea was to name all the template parameters in these `TypeHandler` classes consistently so they can be specified in a more abstract way in the future in the `.toml` files, and potentially carry that into this piece of code. Looking at this now `N` probably needs to be either `T1` or `N0`.

oi/FuncGen.cpp
```diff
@@ -391,3 +430,217 @@ void FuncGen::DeclareGetContainer(std::string& testCode) {
...
+  )";
+
+  constexpr std::string_view pairType = R"(
+    template <typename DataBuffer, typename T1, typename T2>
+    class Pair {
+      public:
+        Pair(DataBuffer db) : _buf(db) {}
+        template <class U>
+        T2 write(U val) {
+          Unit<DataBuffer> second = T1(_buf).write(val);
```
ajor, line 570: Is T1.write() guaranteed to always return a Unit?

Is using a Pair like this explicitly blocked in some way? Pseudo-code:
`Pair<Pair<VarInt, VarInt>, VarInt>`

ajor (commented):
I have a concern about using std::function as it can trigger heap allocations when storing a lambda inside (depending on the size of the lambda and sso-type optimisations).

If I recall correctly, heap allocations are banned in OI's trace code.

Also, std::function is known for being slow, but that's a minor issue that can be resolved at a later date.

ajor:
It looks like we might be able to wrap the lambdas in std::reference_wrappers before putting them into std::function to solve the allocation issue:

https://en.cppreference.com/w/cpp/utility/functional/function/function
> When the target is a function pointer or a [std::reference_wrapper](https://en.cppreference.com/w/cpp/utility/functional/reference_wrapper), small object optimization is guaranteed

It'll also help a little with the performance as std::function will be copying less data into itself (8-byte pointer vs whatever-sized lambda).

oi/FuncGen.cpp
JakeHillion: That type is valid, but calling `.write(...)` on it is blocked at compile time, because it won't return a `Unit`. The expected way to nest pairs is on the right hand side, then the `.write()` drops the top level pair and you're left with a new pair. If you did have a `Pair<Pair<...>, X>` you could instead call `.delegate()` with a callback to complete the left hand side and leave you with the right.

oi/FuncGen.cpp
JakeHillion: Apparently the current method keeps moving the end of the buffer but stops writing anything... I'll replicate that.

JakeHillion:
> I have a concern about using std::function as it can trigger heap allocations when storing a lambda inside (depending on the size of the lambda and sso-type optimisations).
> 
> If I recall correctly, heap allocations are banned in OI's trace code.
> 
> Also, std::function is known for being slow, but that's a minor issue that can be resolved at a later date.

Replaced with a template argument for the lambda, which supposedly solves both of these problems. It'll never heap allocated, and it should be possible to inline the short sections of code which is the main problem with `std::function`.

I think it's going to obfuscate the errors a little, as there is no explicit type declaration for it to moan about and instead a failed template substitution, but it seems the speed/heap allocation are more important.

ajor (approved):

PR #161, opened 2023-06-06
Integration tests: Set up CI testing for TypeGraph

ajor:
CTest can't forward command line arguments to the test runner, so add the option to using an environment variable to enable features instead.

CMake issue tracking the feature that would have been needed:
    https://gitlab.kitware.com/cmake/cmake/-/issues/20470

Tests which aren't passing yet have been disabled in CI.

### DIFF
--- a/test/integration/runner_common.cpp
+++ b/test/integration/runner_common.cpp
@@ -21,14 +21,17 @@ namespace bp = boost::process;
 namespace bpt = boost::property_tree;
 namespace fs = std::filesystem;
 
+bool run_skipped_tests = false;
+
+namespace {
+
 std::string oidExe = OID_EXE_PATH;
 std::string configFile = CONFIG_FILE_PATH;
 
 bool verbose = false;
 bool preserve = false;
 bool preserve_on_failure = false;
-bool run_skipped_tests = false;
-std::vector<std::string> extra_feature_args{};
+std::vector<std::string> global_oid_args{};
 
 constexpr static OIOpts cliOpts{
     OIOpt{'h', "help", no_argument, nullptr, "Print this message and exit"},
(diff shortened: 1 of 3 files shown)


test/integration/runner_common.cpp
```diff
@@ -75,16 +78,17 @@ int main(int argc, char* argv[]) {
...
-        extra_feature_args.push_back(std::string("-f") + optarg);
-        break;
       case 'h':
       default:
         usage(argv[0]);
         return 0;
     }
   }
 
+  if (const char* envArgs = std::getenv("OID_TEST_ARGS")) {
```
JakeHillion, line 88: Could we keep the arg processing too? Should be easy to make the latter one an append and have both. I find the command line stuff easier for manual use on a single test.

test/integration/runner_common.cpp
```diff
@@ -21,13 +21,17 @@ namespace bp = boost::process;
...
+namespace {
+
 std::string oidExe = OID_EXE_PATH;
 std::string configFile = CONFIG_FILE_PATH;
 
 bool verbose = false;
 bool preserve = false;
 bool preserve_on_failure = false;
-bool run_skipped_tests = false;
+std::string global_oid_args{};
```
JakeHillion, line 34: does boost split this correctly if this string contains multiple arguments, or is it expecting one argument per string in the vector?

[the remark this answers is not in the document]
ajor: Done. I've moved the env variable reading higher so that the `--enable-feature` flag still takes priority over the envvar

ajor: It does require one argument per string in the vector, so I've added in a call to `boost::split` for the env var (it didn't work for multiple arguments before)

JakeHillion (approved):
nice.

PR #163, opened 2023-06-06
ci: add include-what-you-use

JakeHillion:
## Summary

Adds the `include-what-you-use` tool to the CI. This ensures that every file includes what it needs and only what it needs.

## Test plan

- CI


JakeHillion:
dropping for now as I can't figure it out. might come back to this later.

PR #167, opened 2023-06-15
Add copyright header to all source files

ajor:
Meta's Open Source Team's automated checks highlighted the missing copyright headers on Headers.cpp.
We had a manual exclusion rule on OITraceCode.cpp, but seeing as how we have core code in this file, I added the header to it and removed the exclusion rule.


JakeHillion:
Thanks for sorting this! My bad for missing `Headers.h`.

JakeHillion (approved):

PR #172, opened 2023-06-20
Flattener: Attempt to take params from parent allocator in case of bad DWARF

ajor:
This is needed for running tests in CI, as the older compilers we have there generate bad DWARF with missing template parameters for allocators.

### DIFF
--- a/oi/type_graph/Flattener.cpp
+++ b/oi/type_graph/Flattener.cpp
@@ -16,6 +16,7 @@
 #include "Flattener.h"
 
 #include "TypeGraph.h"
+#include "TypeIdentifier.h"
 
 namespace type_graph {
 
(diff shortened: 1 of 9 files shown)


ajor:
Reworked to have Flattener do most of the work

oi/type_graph/Flattener.cpp
```diff
@@ -152,6 +191,9 @@ void Flattener::visit(Class& c) {
     flattenParent(parent, flattenedMembers);
   }
 
+  // Perform fixups for bad DWARF
+  fixAllocatorParams(c);
```
JakeHillion, line 195: Would it be nicer to check it's an allocator here? It doesn't feel like very natural to call `fixAllocatorParams` on things which aren't allocators, unless it's a pass of its own.

JakeHillion (approved):

oi/type_graph/Flattener.cpp
ajor: I see what you mean, but I personally have a preference for minimising indentation levels and hiding away complexity.

I was envisioning that in the future there might be various fix-ups we need to do for various kinds of bad DWARF, so we could end up with some code like:
```
fixAllocatorParams(c);
fixSomethingElse(c);
fixAnotherThing(c);
```

Which I think would be cleaner and easier to read at a glance than the alternative:
```
if (...) {
  fixAllocatorParams(c);
}
if (...) {
  fixSomethingElse(c);
}
if (...) {
  fixAnotherThing(c);
}
```

Issue #168, opened 2023-06-16
Remove old feature flags that are duplicated by the feature `-f/-F` enable/disable options


JakeHillion:
I really like this as I have previously said. As you have previously said Jake, we need to remove the other options in conjunction with Strobelight changes.

_Originally posted by @tyroguru in https://github.com/facebookexperimental/object-introspection/pull/125#pullrequestreview-1395327943_
            

Clean up prior use cases of non-feature command line flags to simplify the interface.

PR #174, opened 2023-06-22
features: clean up old flags

JakeHillion:
## Summary

The Features change added a new system of feature flags that don't have a distinct command line argument each. This removes the old flags in favour of the new system.

## Test plan

- `make test`
- CI


JakeHillion:
Closes #168

PR #179, opened 2023-06-23
DrgnParser: Handle enum values in template params

ajor:
We want to use the fully qualified name for scoped enums to keep the C++ compiler happy. When a parameter expects an enum value, we must supply an enum value and not its underlying integer value.

Before:
```
isset_bitset<1, 0>
```

After:
```
isset_bitset<1, apache::thrift::detail::IssetBitsetOption::Unpacked>
```


oi/type_graph/DrgnParser.cpp
```diff
@@ -299,36 +298,67 @@ void DrgnParser::enumerateTemplateParam(drgn_type_template_parameter* tparams,
...
-    } else if (obj->encoding == DRGN_OBJECT_ENCODING_SIGNED) {
-      value = std::to_string(obj->value.svalue);
-    } else if (obj->encoding == DRGN_OBJECT_ENCODING_UNSIGNED) {
-      value = std::to_string(obj->value.uvalue);
-    } else if (obj->encoding == DRGN_OBJECT_ENCODING_FLOAT) {
-      value = std::to_string(obj->value.fvalue);
+
+      drgn_type_enumerator* enumerators = drgn_type_enumerators(obj->type);
+
+      // TODO what about negative enum values or enums with gaps between values?
```
JakeHillion, line 327: It seems like this would SEGV if this was the case, or select a completely wrong value. Can we check this? Even a GLOG `CHECK(enumVal > 0 && enumVal < drgn_type_num_enumerators(obj->type))` would be more helpful if we hit this.

ajor: Good point, thanks - I'd forgotten I'd left this

ajor: Fixed

PR #180, opened 2023-06-23
Timing: add generated code jit timing feature

JakeHillion:
## Summary

Adds a timing call at the beginning and end of the generated code functions and store the time taken in nanoseconds in the data segment. Currently this is output at `LOG(INFO)` level. IMO this belongs in RocksDB and `oid_out.json`, but it requires a bit of a format change so I'll leave it until it's necessary.

This required adding another `uintptr_t` typed element to the data segment prefix, because adding it at the end of the data segment means keeping overhead of setting lengths and such after the call. Probe effect should be minimal as is, but as it's feature gated and off by default I haven't bothered measuring it. I added some names to pieces in these entry functions while I was in there, and also cleaned up a type in the typed data segment entry.

## Test plan
- CI
- `[]: make test: 100% tests passed, 0 tests failed out of 578`
- `["-fjit-timing"]: make test: 100% tests passed, 0 tests failed out of 578`
- `["-ftype-graph"]: make test: 95% tests passed, 31 tests failed out of 578`
- `["-ftype-graph", "-fjit-timing"]: make test: 95% tests passed, 31 tests failed out of 578`
- `["-ftyped-data-segment"]: make test: 87% tests passed, 77 tests failed out of 578`
- `["-ftyped-data-segment", "-fjit-timing"]: make test: 87% tests passed, 77 tests failed out of 578`
(that was hard work)

### DIFF
--- a/oi/FuncGen.cpp
+++ b/oi/FuncGen.cpp
@@ -278,23 +278,38 @@ void FuncGen::DefineTopLevelGetSizeRefTyped(std::string& testCode,
       pointers.initialize();
       pointers.add((uintptr_t)&t);
       auto data = reinterpret_cast<uintptr_t*>(dataBase);
+
+      // TODO: Replace these with types::st::Uint64 once the VarInt decoding
+      // logic is moved out of OIDebugger and into new TreeBuilder.
       data[0] = oidMagicId;
       data[1] = cookieValue;
       data[2] = 0;
+
       size_t dataSegOffset = 3 * sizeof(uintptr_t);
       JLOG("%1% @");
       JLOGPTR(&t);
-      using DataBufferType = OIInternal::TypeHandler<DataBuffer::DataSegment, OIInternal::__ROOT_TYPE__>::type;
-      DataBufferType db = DataBuffer::DataSegment(dataSegOffset);
-      types::st::Unit<DataBuffer::DataSegment> out = OIInternal::getSizeType<DataBuffer::DataSegment>(t, db);
-      types::st::Unit<DataBuffer::DataSegment> final = out.template cast<types::st::Pair<
+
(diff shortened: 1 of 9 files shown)


ajor (approved):
[cppreference](https://en.cppreference.com/w/cpp/chrono/high_resolution_clock) doesn't have a high opinion of `high_resolution_clock`:

> The high_resolution_clock is not implemented consistently across different standard library implementations, and its use should be avoided.
> ...
> Generally one should just use [std::chrono::steady_clock](https://en.cppreference.com/w/cpp/chrono/steady_clock) or [std::chrono::system_clock](https://en.cppreference.com/w/cpp/chrono/system_clock) directly instead of std::chrono::high_resolution_clock: use steady_clock for duration measurements, and system_clock for wall-clock time.

oi/FuncGen.cpp
```diff
@@ -235,29 +235,49 @@ void FuncGen::DefineTopLevelGetObjectSize(std::string& testCode,
...
-      data[0] = oidMagicId;
-      data[1] = cookieValue;
-      data[2] = 0;
 
-      size_t dataSegOffset = 3 * sizeof(uintptr_t);
+      size_t dataSegOffset = 0;
+      data[dataSegOffset++] = oidMagicId;
+      data[dataSegOffset++] = cookieValue;
+      uintptr_t& writtenSize = data[dataSegOffset++];
+      writtenSize = 0;
```
ttreyer, line 261: Would it be simpler to always have `data[3]` dedicated for the timing info?
Set it to 0 when the feature is disabled, populate it otherwise.

ttreyer (changes requested):
It looks mostly good to me 🙂 
Could you change the bool flag to the actual feature set, as discussed, please.
Also take a look at the `data[3]` idea.

ajor (approved):

ttreyer (approved):

PR #183, opened 2023-06-26
feature flag jit-logging

tyroguru:
## Summary
Convert jit-logging to be a feature flag. This removes the `-l` option and adds a `jit-logging` feature flag. This also means that when jit logging is not enabled the JLOG* macros generate nothing so we remove all the extraneous branching.

## Test plan
No new test regressions in the test suite.

I manually ran oid with the current codegen implementation against a test binary (an old version of mttest <sigh>) and the number of instructions generated for the main text segment was 1251 with jit-logging enabled and 885 without (default). Previously this would have been 1251 instructions in both.

I also ran the same oid invocations but with the type-graph implementation (-ftype-graph) and it produced the same results.

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -72,8 +72,9 @@ struct OIArray {
 )";
 }
 
-void defineJitLog(std::string& code) {
-  code += R"(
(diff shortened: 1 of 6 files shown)


JakeHillion (approved):
slick!

Issue #132, opened 2023-04-24
Internal Meta cruft left in trace code


tyroguru:
Some internal Meta cruft is hanging around in OITraceCode.cpp which needs removing.

https://github.com/facebookexperimental/object-introspection/blob/4a64fc5c9c57aa2d2f9856f4d9b66d5148c83eab/src/OITraceCode.cpp#L93

Testing with the specific object should be done.

tyroguru:
Closed with #186

PR #186, opened 2023-06-27
remove unused old code from trace code

tyroguru:
## Summary
Removes some bits from tracing code that are no longer needed. This addresses issue #132 .

@arsarwade says that `AllocAligned` and `AlignedDeleter` are both vestiges of how we used to handle template parameters with custom allocators so can go. @ajor, I added you as a reviewer to check you're happy with that and have no need for them.
 
## Test plan

I have successfully codegen'd an `Adunit` and and an `AdIndexerUnit`. No new test failures are seen.

### DIFF
--- a/oi/OITraceCode.cpp
+++ b/oi/OITraceCode.cpp
@@ -106,55 +106,6 @@ void __jlogptr(uintptr_t ptr) {
 
 }  // namespace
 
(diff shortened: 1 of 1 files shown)


ajor (approved):
No objections from me!

PR #187, opened 2023-06-27
llvm: update to llvm/clang 15

JakeHillion:
## Summary

Updates our LLVM/Clang libraries to 15. This improves the behaviour of the new ORC linker stack, used in some of the Typed TreeBuilder work.

I let the CI build with clang-15 now too, and it raised a bunch of new warnings. These are all fixed in this PR too.

## Test plan

- CI
- `make test`

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,8 +43,8 @@ workflows:
 
       - build:
           name: build-clang
-          cc: /usr/bin/clang-12
-          cxx: /usr/bin/clang++-12
+          cc: /usr/bin/clang-15
+          cxx: /usr/bin/clang++-15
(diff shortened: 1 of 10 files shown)


JakeHillion:
These problems seem to be related to using clang-15 to compile and not using the clang-15/llvm-15 libraries. This can be seen by the GCC CI tests passing. Will come back to this and debug it further tomorrow.

JakeHillion:
Reproduced the clang-15 compiler causing a problem on my devserver. It's important that this gets fixed so will hold off on landing this until I've figured out why.

JakeHillion:
Closing for #342 which is synced with intern

PR #190, opened 2023-06-28
Take list of pass-through types from config instead of hardcoding

ajor:
As we now store ContainerInfo objects in OICodeGen::Config, we can not copy it any more. Change all places that took copies to take const references instead.

The copy in OICodeGen modified membersToStub, the contents of which form part of OICache's hash. However, as OICache also previously had its own copy, it would not have been OICodeGen's modifications.

### DIFF
--- a/dev.oid.toml
+++ b/dev.oid.toml
@@ -47,3 +47,8 @@ containers = [
     "PWD/types/thrift_isset_type.toml",
     "PWD/types/weak_ptr_type.toml",
 ]
+pass_through = [
+    ["std::allocator", "memory"],
+    ["std::char_traits", "string"],
+    ["folly::fbstring_core", "folly/FBString.h"],
+]
--- a/oi/OICodeGen.h
+++ b/oi/OICodeGen.h
@@ -51,18 +51,19 @@ struct ParentMember {
 class OICodeGen {
  public:
   struct Config {
-    /*
-     * Note: don't set default values for the config so the user gets an
-     * uninitialized field" warning if they missed any.
-     */
+    Config() = default;
+    Config(const Config& other) = delete;
+    Config& operator=(const Config& other) = delete;
+    Config(Config&& other) = delete;
+    Config& operator=(Config&& other) = delete;
+
     bool useDataSegment;
-
-    FeatureSet features{};
-
-    std::set<fs::path> containerConfigPaths{};
-    std::set<std::string> defaultHeaders{};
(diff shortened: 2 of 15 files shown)


oi/OICodeGen.h
```diff
@@ -51,18 +51,25 @@ struct ParentMember {
...
     bool useDataSegment;
 
-    FeatureSet features{};
+    FeatureSet features;
 
-    std::set<fs::path> containerConfigPaths{};
-    std::set<std::string> defaultHeaders{};
-    std::set<std::string> defaultNamespaces{};
-    std::vector<std::pair<std::string, std::string>> membersToStub{};
+    std::set<fs::path> containerConfigPaths;
```
ajor, line 68: The `{}` makes no difference on these members, as class-objects are always default initialised if not initialised explicitly.

However, the error messages are a lot clearer without the `{}` and removing them allowed me to track down all the locations OICodeGen was being copied.

oi/OICodeGen.h
```diff
@@ -51,18 +51,25 @@ struct ParentMember {
...
+    Config() = default;
+    Config(const Config& other) = delete;
+    Config& operator=(const Config& other) = delete;
+    Config(Config&& other) = default;
+    Config& operator=(Config&& other) = default;
+
     /*
      * Note: don't set default values for the config so the user gets an
      * uninitialized field" warning if they missed any.
      */
```
ajor, line 63: @ttreyer I don't understand what exactly this comment is getting at. I've tried a few things and can't seem to get an uninitialised field warning

dev.oid.toml
```diff
@@ -47,3 +47,8 @@ containers = [
     "PWD/types/thrift_isset_type.toml",
     "PWD/types/weak_ptr_type.toml",
 ]
+pass_through = [
```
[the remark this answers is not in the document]
ajor: This file isn't documented yet.

We've got READMEs for the [container types config format](https://github.com/facebookexperimental/object-introspection/blob/main/types/README.md) and for the [integration test format](https://github.com/facebookexperimental/object-introspection/blob/main/test/integration/README.md)

tyroguru (approved):
That's great Alastair. Just one point about documenting the new option which I'm sure you've got covered anyway.

ajor:
Thanks, yeah we will document this file at some point, but I'll leave it for a separate PR

PR #194, opened 2023-06-29
TopoSorter: Fix sorting of container template parameters

ajor:
For std::vector and std::list, template parameters are not required to be defined before they can be used. Delay sorting them until the end.

Also fix a CodeGen bug where we were defining typedefs in the middle of the forward declarations. They only need to be defined when other types are defined.

References:
https://eel.is/c++draft/vector.overview#4
http://eel.is/c++draft/list.overview#3
https://eel.is/c++draft/forward.list.overview#4

### DIFF
--- a/oi/type_graph/TopoSorter.cpp
+++ b/oi/type_graph/TopoSorter.cpp
@@ -55,29 +55,57 @@ void TopoSorter::visit(Type& type) {
 }
 
 void TopoSorter::visit(Class& c) {
-  for (const auto& param : c.templateParams) {
-    visit(param.type);
-  }
   for (const auto& parent : c.parents) {
     visit(*parent.type);
   }
   for (const auto& mem : c.members) {
     visit(*mem.type);
   }
-  sortedTypes_.push_back(c);
-
-  // Same as pointers, child do not create a dependency so are delayed until the
-  // end
-  for (const auto& child : c.children) {
-    typesToSort_.push(child);
-  }
-}
-
-void TopoSorter::visit(Container& c) {
   for (const auto& param : c.templateParams) {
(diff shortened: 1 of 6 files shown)


oi/type_graph/TopoSorter.cpp
```diff
@@ -54,30 +54,68 @@ void TopoSorter::visit(Type& type) {
...
   }
+  for (const auto& param : c.templateParams) {
+    visit(param.type);
+  }
   sortedTypes_.push_back(c);
 
-  // Same as pointers, child do not create a dependency so are delayed until the
-  // end
+  // Same as pointers, children do not create a dependency so are delayed until
+  // the end TODO comment
```
ajor, line 80: Finish this off!

tyroguru: Looks great and I couldn't see anything TBH.

One general point. Don't be afraid to put comments in so that anyone coming into this code can readily pick this up. It's a small thing but I do like at least one comment that gives an overview of what the code in a file is trying to do without the reader having to ingest it all. Not for now but just to bare in mind.

tyroguru (approved):

oi/type_graph/TopoSorter.cpp
ajor: There is a short overview comment in the header file: https://github.com/facebookexperimental/object-introspection/blob/5560624c0b5301f8b5fb0eb955bffc0560f8e384/oi/type_graph/TopoSorter.h#L29

The `visitAfter` function is probably the most confusing bit (as it's not part of a standard topological sort) and is conviently the part I only half wrote a comment for 😅

PR #196, opened 2023-06-29
integration: add GLOB with CONFIGURE_DEPENDS to pick up toml files

JakeHillion:
## Summary

Currently integration test configs have to be specified manually in the `CMakeLists.txt` file. This changes to using a CMake `GLOB_RECURSE` with the flag [`CONFIGURE_DEPENDS`](https://cmake.org/cmake/help/latest/command/file.html#glob-recurse) (available since 3.12) which causes the build to check the glob each time and re-configure CMake if it has changed.

## Test plan

```
$ cp test/integration/std_string{,_2}.toml && make test
[0/2] Re-checking globbed directories...
-- GLOB mismatch!
[1/2] Re-running CMake...
...
```

Otherwise:
```
$ make test
[0/2] Re-checking globbed directories...
[0/11] Building drgn
...
```

### DIFF
--- a/test/integration/CMakeLists.txt
+++ b/test/integration/CMakeLists.txt
@@ -1,65 +1,22 @@
-# Add new test definition files to this list:
-set(INTEGRATION_TEST_CONFIGS
-  alignment.toml
-  anonymous.toml
-  arrays.toml
-  cycles.toml
-  enums.toml
-  enums_params.toml
-  fbstring.toml
-  ignored.toml
-  inheritance_access.toml
-  inheritance_multiple.toml
-  inheritance_polymorphic.toml
-  inheritance_polymorphic_diamond.toml
-  inheritance_polymorphic_non_dynamic_base.toml
-  multi_arg.toml
-  namespaces.toml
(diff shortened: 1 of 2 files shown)


ajor (approved):
The "Adding Tests" section of the docs will need updating: https://github.com/facebookexperimental/object-introspection/blob/b50ab57e3307a576e2692bce420e7791ec1a2508/test/integration/README.md?plain=1#L33

Just quoting this section from the CMake docs for future reference:
> Note We do not recommend using GLOB to collect a list of source files from your source tree. If no CMakeLists.txt file changes when a source is added or removed then the generated build system cannot know when to ask CMake to regenerate. The CONFIGURE_DEPENDS flag may not work reliably on all generators, or if a new generator is added in the future that cannot support it, projects using it will be stuck. Even if CONFIGURE_DEPENDS works reliably, there is still a cost to perform the check on every rebuild.

They recommend against this method, but it seems to be fairly widely used and works with our current generators. We can always switch back if we encounter problems in the future.

PR #206, opened 2023-07-05
TypeGraph: Make Primitive types singletons

ajor:
Should save some memory and improve performance by not having loads of copies of primitive types.


ajor:
Updated new function name to match the existing `make_type` so that it acts as an override, as intended. Also added a static assert to block compilation if this singleton override isn't used.

tyroguru (approved):

PR #209, opened 2023-07-05
add drgn aranges & non-debug patches

JakeHillion:
## Summary

Backports @ttreyer's drgn fixes.


ajor (approved):
🤷‍♂️

Approving because the drgn commit titles are so reassuring

PR #216, opened 2023-07-07
create strict mode and enable it for tests

JakeHillion:
## Summary

Currently tests do not fail if TreeBuilder did not completely consume the data segment and they definitely should. I think there's a solid argument to be made for making this fatal everywhere, but TreeBuilder V2 should remove this possibility anyway.

This PR creates a `strict` mode that causes TreeBuilder to fatally error if it hasn't read all of the data segment and enables it for tests. This exposes a few issues under Type Graph/Typed Data Segment that will need fixing first.

I was planning to hold off landing this until the bugs were fixed, but I think it makes more sense to blacklist the failing tests on the CI as we already do that with some. All tests are passing without `-ftype-graph` or `-ftyped-data-segment`.

## Test plan

- CI


ajor (approved):
Very useful!

PR #217, opened 2023-07-07
TypeGraph: Rename visit(Type) functions to accept(Type)

ajor:
`visit(Type&)` and `visit(Type*)` were helper functions than did cycle detection and provided nicer syntax for the `type.accept(*this)` calls.

However, because they overloaded the `visit()` function, it was easy to accidentally call a concrete visit method (e.g. `visit(Class&)`) instead of the virtual-dispatching `visit(Type&)`.

By changing the name of this wrapper, it will make it much more obvious when code is introduced which bypasses the cycle detection.


JakeHillion (approved):
love making overloads explicit, nice!

PR #219, opened 2023-07-07
NameGen: Remove invalid characters from member names

ajor:
GCC includes dots in vptr member names, e.g. "_vptr.MyClass". These aren't valid in C++, so we must replace them.

### DIFF
--- a/oi/type_graph/NameGen.cpp
+++ b/oi/type_graph/NameGen.cpp
@@ -66,6 +66,10 @@ void NameGen::visit(Class& c) {
   // Deduplicate member names. Duplicates may be present after flattening.
   for (size_t i = 0; i < c.members.size(); i++) {
     c.members[i].name += "_" + std::to_string(i);
+
+    // GCC includes dots in vptr member names, e.g. "_vptr.MyClass"
+    // These aren't valid in C++, so we must replace them
+    std::replace(c.members[i].name.begin(), c.members[i].name.end(), '.', '$');
   }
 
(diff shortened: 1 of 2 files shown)


oi/type_graph/NameGen.cpp
```diff
@@ -66,6 +66,10 @@ void NameGen::visit(Class& c) {
   // Deduplicate member names. Duplicates may be present after flattening.
   for (size_t i = 0; i < c.members.size(); i++) {
     c.members[i].name += "_" + std::to_string(i);
+
+    // GCC includes dots in vptr member names, e.g. "_vptr.MyClass"
+    // These aren't valid in C++, so we must replace them
+    std::replace(c.members[i].name.begin(), c.members[i].name.end(), '.', '$');
```
JakeHillion, line 72: could we make this a general function somewhere? if we made it a static method on `NameGen` it could be used elsewhere and I think I need something similar in my unfinished cycles fix

JakeHillion (approved):

oi/type_graph/NameGen.cpp
ajor: A function to replace characters in names? Isn't the NameGen pass working for your use-case?

ajor: I think anything to do with ensuring names are valid should be the responsibility of NameGen - we can just run your new pass before NameGen and have it fix up any problematic names as well as perform the deduplication

PR #221, opened 2023-07-10
comment existing typed data segment work

JakeHillion:
## Summary

Comment the logic for `-ftyped-data-segment`. Trivial functional changes.

## Test plan

- `make test`
- CI

### DIFF
--- a/include/oi/types/st.h
+++ b/include/oi/types/st.h
@@ -16,10 +16,50 @@
 #ifndef OI_TYPES_ST_H
 #define OI_TYPES_ST_H 1
 
-namespace ObjectIntrospection {
-namespace types {
-namespace st {
+/*
+ * Static Types
+ *
+ * OI employs a data segment to transfer information about the probed object to
+ * the debugger. Static Types are used with the `-ftyped-data-segment` feature
+ * to provide a compile time description of the contents of this data segment.
+ *
+ * DataBuffer represents any type with two methods: `void write_byte(uint8_t)`,
+ * which writes a given byte to the buffer; and, `size_t offset()`, which
+ * returns the number of bytes written. Each Static Type holds a DataBuffer
+ * which describes where to write data, and has no other fields. DataBuffers
+ * should remain pointer sized enabling trivial copies.
+ *
+ * Writing to an object of a given static type returns a different type which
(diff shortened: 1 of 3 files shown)


include/oi/types/st.h
```diff
@@ -16,10 +16,52 @@
...
+ * OI employs a data segment to transfer information about the probed object to
+ * the debugger. Static Types are used with the `-ftyped-data-segment` feature
+ * to provide a compile time description of the contents of this data segment.
+ *
+ * DataBuffer represents any type with two methods: `void write_byte(uint8_t)`,
+ * which writes a given byte to the buffer; and, `size_t offset()`, which
+ returns
+ * the number of bytes written. The only implementation at present is
+ DataSegment,
+ * which is defined in `FuncGen::DefineDataSegmentDataBuffer`. Each Static Type
```
ajor, line 31:
```suggestion
 * returns the number of bytes written. The only implementation at present is
 * DataSegment, which is defined in `FuncGen::DefineDataSegmentDataBuffer`. Each
 * Static Type ...
```

A classic formatting nitpick ;)

I'm guessing the linter wrapped your lines poorly. This has happened in a few lines in this block comment, not just here.

include/oi/types/st.h
```diff
@@ -16,10 +16,52 @@
...
+ * Static Types
+ *
+ * OI employs a data segment to transfer information about the probed object to
+ * the debugger. Static Types are used with the `-ftyped-data-segment` feature
+ * to provide a compile time description of the contents of this data segment.
+ *
+ * DataBuffer represents any type with two methods: `void write_byte(uint8_t)`,
+ * which writes a given byte to the buffer; and, `size_t offset()`, which
+ returns
+ * the number of bytes written. The only implementation at present is
```
ajor, line 29: Just a warning: if you're going to say "the only implementation at present" then we'll have to be sure to update this comment if that ever changes. (are you planning to change add more?)

It might be more future-proof to be less specific here, but wording is just a matter of preference so feel free to leave it as-is if you like.

ajor (approved):
Great comments! I thought I understood this stuff before, but it's much clearer in my mind now after reading through these!

PR #229, opened 2023-07-11
EnforceCompatibility: Stub out void pointers

ajor:
CodeGen v1 does not record anything for pointers to incomplete types. Not even the address, as is done for other pointers.

Introduce a new Primitive type "Incomplete". This behaves identically to "Void", but allows us to tell whether a type was defined as void or if it ended up like that because of incomplete DWARF information.


ajor:
Looking back on this code now again, I'm not sure if "EnforceCompatibility" is the right pass for it - maybe it should be part of "RemoveMembers" instead?

EnforceCompatibility is for behaviour to produce the same results as CodeGen v1, but that we plan to change in the future.
RemoveMembers is for behaviour that we expect to keep permanently.

It depends on whether we'd want to record the pointer addresses for incomplete types in the future or have this stubbing behaviour be permanent. Could the addresses be useful?

Thoughts?

JakeHillion:
> Looking back on this code now again, I'm not sure if "EnforceCompatibility" is the right pass for it - maybe it should be part of "RemoveMembers" instead?
> 
> EnforceCompatibility is for behaviour to produce the same results as CodeGen v1, but that we plan to change in the future. RemoveMembers is for behaviour that we expect to keep permanently.
> 
> It depends on whether we'd want to record the pointer addresses for incomplete types in the future or have this stubbing behaviour be permanent. Could the addresses be useful?
> 
> Thoughts?

My thoughts on this were that you were planning to change it with v2. That is, record the pointer for something incomplete in the same way it records a void pointer. IMO it makes sense to treat them the same way as `void*`, because they both represent an opaque pointer.

The question of whether it makes sense to record void pointers at all is a different one. We probably record far too many pointers that aren't useful at the minute, at least in my experience. I'm wondering whether this is a holdover from when TreeBuilder did cycle detection, and now that it's fully in the JIT code it might be unnecessary.

Either way, I think this makes sense in `EnforceCompatibility` as we should end up treating `void*` and `class T*` the same. Whether we should record their pointers at all is a different question, and also about future behaviour instead of compatible behaviour.

JakeHillion (approved):

PR #230, opened 2023-07-11
Flattener: Handle incomplete parent types

ajor:


JakeHillion:
I'm a little confused by this change. Where does an incomplete parent show up, is it valid C++ or bad DWARF? Why is it okay to act as if the parent is not there during the flattener in the latter case, does it not mess up offsets? Some comments in the code would be really useful.

ajor:
It's bad DWARF (spotted by @arsarwade in his experiments) - I've expanded the comment to explain this.

The gap caused by the missing parent is a more general issue that appears throughout the TypeGraph passes, e.g. we also get gaps when stubbing unions or using the `codegen.ignore` config option. The `AddPadding` pass is run after other transformations and is responsible for fixing up any gaps: https://github.com/facebookexperimental/object-introspection/blob/ececbe0a99f519cc8f2ba60b898b1e679a8f4a9a/oi/CodeGen.cpp#L840-L842

JakeHillion:
> It's bad DWARF (spotted by @arsarwade in his experiments) - I've expanded the comment to explain this.
> 
> The gap caused by the missing parent is a more general issue that appears throughout the TypeGraph passes, e.g. we also get gaps when stubbing unions or using the `codegen.ignore` config option. The `AddPadding` pass is run after other transformations and is responsible for fixing up any gaps:
> 
> https://github.com/facebookexperimental/object-introspection/blob/ececbe0a99f519cc8f2ba60b898b1e679a8f4a9a/oi/CodeGen.cpp#L840-L842

Of course, I forgot that we rely on the alignments of the members to sort this later. Thanks for clearing that up.

JakeHillion (approved):

PR #231, opened 2023-07-12
type_graph: add CycleFinder pass

JakeHillion:
## Summary

Typed Data Segment can't currently handle cycles, because it's impossible to define the cyclic type in unmodified template parameters. This change solves that by pulling out the graph edges which would be problematic first, creating a fake type which can be used in place of the cyclic type.

This implementation corrects cycles in a subset of cases, but should expose all cycles. This allows it to give a descriptive error to continue debugging.

## Test plan

- `OID_TEST_ARGS='-ftyped-data-segment' make test` - all cycle failures are gone.

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -28,6 +28,7 @@
 #include "type_graph/AddChildren.h"
 #include "type_graph/AddPadding.h"
 #include "type_graph/AlignmentCalc.h"
+#include "type_graph/CycleFinder.h"
 #include "type_graph/DrgnParser.h"
 #include "type_graph/Flattener.h"
 #include "type_graph/NameGen.h"
@@ -40,6 +41,7 @@
 
 using type_graph::Class;
 using type_graph::Container;
+using type_graph::CycleBreaker;
 using type_graph::Enum;
 using type_graph::Member;
 using type_graph::Type;
@@ -126,6 +128,12 @@ void addIncludes(const TypeGraph& typeGraph,
   }
 }
 
+void genDeclsCycleBreaker(const CycleBreaker& b, std::string& code) {
+  code += "struct ";
+  code += b.name();
+  code += ";\n";
+}
+
 void genDeclsClass(const Class& c, std::string& code) {
   if (c.kind() == Class::Kind::Union)
     code += "union ";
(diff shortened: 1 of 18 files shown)


JakeHillion:
Changing to draft as I'm having a bit of a rethink of this. It might end up not changing but draft for now.

JakeHillion:
Waiting for a rebase then I'll clean this up.

oi/CodeGen.cpp
```diff
@@ -126,6 +128,12 @@ void addIncludes(const TypeGraph& typeGraph,
   }
 }
 
+void genDeclsCycleBreaker(const CycleBreaker& b, std::string& code) {
```
ajor, line 131: Does CycleBreaker not need a generated definition too?

oi/type_graph/Printer.cpp
```diff
@@ -132,6 +132,14 @@ void Printer::visit(const DummyAllocator& d) {
   print(d.allocType());
 }
 
+void Printer::visit(const CycleBreaker& b) {
+  if (prefix(&b))
+    return;
+
+  out_ << "CycleBreaker" << std::endl;
```
ajor, line 139: If we're going to generate a name for CycleBreaker nodes then we should print it out as well - Printer is meant to be able to show every detail about a type graph. I do plan to move the NameGen tests over to the Printer-based testing framework at some point too.

oi/type_graph/NameGen.cpp
```diff
@@ -152,4 +152,28 @@ void NameGen::visit(Typedef& td) {
   accept(td.underlyingType());
 }
 
+void NameGen::visit(CycleBreaker& b) {
+  accept(b.to());
+
+  std::string name = "fake_" + b.to().name();
```
ajor, line 158: Following on from the comment on Printer - is it actually useful to have names for CycleBreaker nodes at all or could they just be handled like Dummy nodes, where they're just templated on the "to" type, e.g. `CycleBreaker<MyType>`

oi/type_graph/TopoSorter.cpp
```diff
@@ -117,6 +117,11 @@ void TopoSorter::visit(Typedef& td) {
   sortedTypes_.push_back(td);
 }
 
+void TopoSorter::visit(CycleBreaker& b) {
```
ajor, line 120: Could we get a TopoSorterTest for this please?

ajor (commented):
Would we be able to get a few more unit tests for CycleBreaker in here? It's letting the coverage side down! https://app.codecov.io/gh/facebookexperimental/object-introspection/pull/231/tree/oi/type_graph

JakeHillion:
Partially addresses #293

JakeHillion:
Closing this PR as it's extremely outdated and a mess. Some of the approach is still okay.

PR #236, opened 2023-07-14
type_graph: deduplicate has_node_id

JakeHillion:
## Summary

The check for whether to require a node ID is completed in two places, neither of which are at the type which describes whether it needs a node ID. Move this boolean to within the types for clarity/deduplication.

## Test plan

- CI


ajor (approved):
Looks good, just one point on consistency:

This adds `has_node_id` to two non-Type nodes `Member` and `TemplateParam`. It hasn't been added to the other two non-Type nodes though: `Function` and `Parent`.

I don't think it matters if they have this field or not, as long as they're all the same.

PR #237, opened 2023-07-14
ci: order coverage jobs

JakeHillion:
## Summary

Places ordering constraints on the four coverage jobs that run in CI. Recently I've seen a lot of 502 errors where codecov requests you try again in 30 seconds. Hopefully making sure these jobs aren't concurrent will help with that.

## Test plan

- CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -43,14 +43,17 @@ workflows:
           name: coverage-type-graph
           requires:
             - test-type-graph-gcc
+            - coverage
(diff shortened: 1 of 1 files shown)


ajor (approved):

JakeHillion:
The new dependencies cause circleci to load both workspaces, which causes the same file collision problem. Need to do this properly instead and have the paths differentiated such that one coverage job can load the details from all the tests.

PR #238, opened 2023-07-14
Cyclic containers

JakeHillion:
## Summary

Extends support from just raw pointers to container template arguments.

I was expecting to have to make the intermediate type a proper dummy (size & alignment) to work with containers, but all the tests pass with it simply declared. I now think this is fine, as all cycles must show up as pointers/references at some point. That is, we shouldn’t claim that a Vector of Foo is is causing a cycle (Vector -> Foo specifically) because that’s the owner. `Vector<Foo*>` / `Vector<std::reference_wrapper<Foo>>` are both fine as the actual cycle is caused by a pointer/reference and not by the vector itself, and thus size/alignment will be fine.

This is a little rambly but I’ve struggled to reason about where these cycles may occur and concluded this is fine.

## Test plan
- CI


JakeHillion:
collapsed into #231

PR #243, opened 2023-07-18
TypeIdentifier: Stub unions

ajor:
In general, we can't tell which member is active in a union so it is not safe to try and measure any of them.


JakeHillion (approved):
Clean.

arsarwade:
Could you please check that alignment of union is followed. I think we do it today in old codegen.

Basically if 

```
union {char a; uint64_t b;}
```

We need to make sure when the union is padded out, that it has an alignment of 8 in the above case. Otherwise some cases where alignment matters can start to break.

ajor:
Yeah, I just discovered that this change does break union alignment. I'm adding a test and fixing this now

ajor:
Converted back to a draft as the new approach which handles alignment properly now depends on a stack of other PRs:
#247
#248
#249

ajor:
This PR is back, now including an `alignas` specifier for unions and with tests for union alignment already checked in: https://github.com/facebookexperimental/object-introspection/blob/2d1cc92bb4935d5eb44edb142571349f30ba448c/test/integration/unions.toml#L57

PR #249, opened 2023-07-19
Rename RemoveIgnored -> RemoveMembers

ajor:
Also reshuffle CodeGen's passes to fix an alignment bug with removed members.

Change RemoveMembers to actually remove members instead of replacing them with padding. AddPadding must be run afterwards to fill in the gaps.

### DIFF
--- a/oi/type_graph/RemoveMembers.cpp
+++ b/oi/type_graph/RemoveMembers.cpp
@@ -13,26 +13,26 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "RemoveIgnored.h"
+#include "RemoveMembers.h"
 
 #include "AddPadding.h"
 #include "TypeGraph.h"
 
 namespace type_graph {
 
-Pass RemoveIgnored::createPass(
+Pass RemoveMembers::createPass(
     const std::vector<std::pair<std::string, std::string>>& membersToIgnore) {
   auto fn = [&membersToIgnore](TypeGraph& typeGraph) {
-    RemoveIgnored removeIgnored{typeGraph, membersToIgnore};
+    RemoveMembers removeMembers{membersToIgnore};
     for (auto& type : typeGraph.rootTypes()) {
-      removeIgnored.accept(type);
+      removeMembers.accept(type);
     }
   };
 
-  return Pass("RemoveIgnored", fn);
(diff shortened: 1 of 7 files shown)


oi/type_graph/RemoveMembers.cpp
```diff
@@ -54,19 +54,12 @@ void RemoveIgnored::visit(Class& c) {
...
-      continue;
-    }
-    auto& primitive = typeGraph_.makeType<Primitive>(Primitive::Kind::Int8);
-    auto& paddingArray =
-        typeGraph_.makeType<Array>(primitive, c.members[i].type().size());
-    c.members[i] =
-        Member{paddingArray, AddPadding::MemberPrefix, c.members[i].bitOffset};
-  }
+  std::erase_if(c.members, [this, &c](Member member) {
+    return ignoreMember(c.name(), member.name);
```
JakeHillion, line 58: Given `c.name()` returns a copy, can you gather it before and capture it by reference?

oi/type_graph/RemoveMembers.cpp
```diff
@@ -54,19 +54,12 @@ void RemoveIgnored::visit(Class& c) {
...
+  });
 }
 
-bool RemoveIgnored::ignoreMember(const std::string& typeName,
+bool RemoveMembers::ignoreMember(const std::string& typeName,
```
JakeHillion, line 62: Unrelated but nearby, why is this not checking a set? It seems better to implement a hash function for `std::pair<std::string, std::string>` than loop over a vector of pairs. Even a map of ContainerName -> Set<FieldName> would be a lot better.

JakeHillion (approved):

oi/type_graph/RemoveMembers.cpp
ajor: The real reason: It's sharing the implementation with CodeGen v1 and I didn't want to change too much.

But in terms of efficiency, I think the winner of vectors vs maps/sets would come down to how many elements are typically being stored. For a small number of elements a vector would likely win due to the lack of hashing and being more cache friendly.

oi/type_graph/RemoveMembers.cpp
ajor: No more copy: #264

PR #250, opened 2023-07-19
AlignmentCalc fixes

ajor:
- Recurse into params,parents,members,children of Classes
- Extend TypeGraphParser to understand bitfields
- Add tests


JakeHillion (approved):

ajor:
Added one more fix: Array alignment should be based on elementType's alignment

PR #252, opened 2023-07-20
googletest: change to FetchContent

JakeHillion:
## Summary

The internal version of GoogleTest we were using has disappeared and the CentOS version is old. Vendor it with FetchContent.

## Test plan

- CI
- `make clobber; make configure-devel && make test`
- `make clobber; make configure-static && make test`
Static is a complete mess, and there are a tonne of test failures on it too. I think this is pre-existing though?

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -54,6 +54,16 @@ target_compile_options(stl_logging_unittest PRIVATE "-w")
 target_compile_options(symbolize_unittest PRIVATE "-w")
 target_compile_options(utilities_unittest PRIVATE "-w")
 
+### googletest
+# Do this in the main file so it can be fetched before setting project warnings.
+# After this is fixed with FetchContent, move to test/CMakeLists.txt.
+FetchContent_Declare(
+  googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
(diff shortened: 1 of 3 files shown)


CMakeLists.txt
```diff
@@ -54,6 +54,16 @@ target_compile_options(stl_logging_unittest PRIVATE "-w")
 target_compile_options(symbolize_unittest PRIVATE "-w")
 target_compile_options(utilities_unittest PRIVATE "-w")
 
+### googletest
+# Do this in the main file so it can be fetched before setting project warnings.
+# After this is fixed with FetchContent, move to test/CMakeLists.txt.
+FetchContent_Declare(
+  googletest
+  GIT_REPOSITORY https://github.com/google/googletest.git
+  GIT_TAG        1ed6a8c67a0bd675149ece27bbec0ef1759854cf
```
ajor, line 63: Might be safer to go with the hash of a released gtest version instead of the latest commit on main?

ajor (approved):

CMakeLists.txt
JakeHillion: I went with this due to https://github.com/google/googletest#live-at-head which makes sense to me (though I wonder how often we'd update it...). Happy to be vetoed though.

PR #253, opened 2023-07-21
provide common interfaces for continue and detach requests

tyroguru:
## Summary
The `OIDebugger` class provides a public interface for oid to detach or continue the main target thread via `PTRACE_DETACH` and `PTRACE_CONT` requests respectively. The `OIDebugger` implementation has a number of call sites which continue a thread but it may not be the main thread (`traceePid`) and we may need to pass a signal to inject via the `data` parameter. This change simply tidies all these individual call sites under a common private implementation. The public interface has been converted to use the use the private interface for its close requests.

There are a few PTRACE_DETACH request sites in `OIDebugger.cpp` and I'll tify them up into a common interface in another change.

## Test plan
A `make test` has no new failures.

### DIFF
--- a/oi/OIDebugger.cpp
+++ b/oi/OIDebugger.cpp
@@ -445,23 +445,50 @@ bool OIDebugger::isExtendedWait(int status) {
   return (getExtendedWaitEventType(status) != 0);
 }
 
+/*
+ * Continue the main traced thread either by a detach request (default)
+ * or via a continue. No provision is made for passing a signal.
+ */
 bool OIDebugger::contTargetThread(bool detach) const {
-  VLOG(4) << "contTargetThread: About to PTRACE_CONT pid " << std::dec
-          << traceePid << " detach = " << detach;
+  VLOG(4) << "contTargetThread: About to " << (detach ? "detach" : "continue")
+          << " pid " << std::dec << traceePid;
 
-  if (detach) {
-    if ((ptrace(PTRACE_DETACH, traceePid, nullptr, nullptr)) < 0) {
-      LOG(ERROR) << "contTargetThread failed to detach pid " << traceePid << " "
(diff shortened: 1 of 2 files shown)


oi/OIDebugger.cpp
```diff
@@ -445,23 +445,54 @@ bool OIDebugger::isExtendedWait(int status) {
...
 bool OIDebugger::contTargetThread(bool detach) const {
-  VLOG(4) << "contTargetThread: About to PTRACE_CONT pid " << std::dec
-          << traceePid << " detach = " << detach;
+  bool ret = true;
+
+  VLOG(4) << "contTargetThread: About to " << (detach ? "detach" : "continue")
+          << " pid " << std::dec << traceePid;
 
+  errno = 0;
   if (detach) {
```
JakeHillion, line 459: Could this be cleaned up a bit more? I'm thinking:
```
if (!detach)
  return contTargetThread(traceePid);

// detach logic, returning at the end
```

oi/OIDebugger.cpp
```diff
@@ -445,23 +445,54 @@ bool OIDebugger::isExtendedWait(int status) {
...
+  bool ret = true;
+
+  VLOG(4) << "contTargetThread: About to continue pid " << std::dec
+          << targetPid;
+
+  errno = 0;
+  if ((ptrace(PTRACE_CONT, targetPid, nullptr, data)) < 0) {
+    LOG(ERROR) << "contTargetThread failed to continue pid " << targetPid << " "
+               << strerror(errno) << "(" << errno << ")";
+    ret = false;
```
JakeHillion, line 492: s/`ret = false;`/`return false;`/

JakeHillion (approved):

PR #255, opened 2023-07-22
Improve handling of newly created threads in initial seize code

tyroguru:
## Summary
Our initial thread seize code is very racy and it's super easy to end up with threads being ignored or hung for long periods waiting for stop states to be reaped. This change aims to handle the case where a thread that has been seized executes a sequence such that a new thread is created (e.g., a clone(2) syscall and the newborn thread shows up in the /proc namespace so the directory_iterator picks it up. This new thread will be traced by us already as the tracing sate is inherited from its parent thread so we can't seize it. It will however be in a signal-stop state so that should be reaped and the thread continued.

## Test plan
make 'test' returns the same failures as before the change. However, it is unlikely to show any change as this code is handling an `EPERM` return from `ptrace(2)`. Test in prod really. I have engineered highly controlled test cases to work through the failure scenarios and at a minimum I don't think we are any more exposed to failure with this code than without.

### DIFF
--- a/oi/OIDebugger.cpp
+++ b/oi/OIDebugger.cpp
@@ -123,26 +123,34 @@ bool OIDebugger::patchFunctions(void) {
  * Single step an instruction in the target process 'pid' and leave the target
  * thread stopped. Returns the current rip.
  */
-uint64_t OIDebugger::singlestepInst(pid_t pid, struct user_regs_struct& regs) {
+uint64_t OIDebugger::singleStepInst(pid_t pid, struct user_regs_struct& regs) {
   int status = 0;
 
   Metrics::Tracing _("single_step_inst");
 
+  errno = 0;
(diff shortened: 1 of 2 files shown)


oi/OIDebugger.cpp
```diff
@@ -968,7 +980,7 @@ OIDebugger::processTrapRet OIDebugger::processTrap(pid_t pid,
   siginfo_t info;
 
   auto stopsig = WSTOPSIG(status);
-  VLOG(4) << "Stop sig: " << std::dec << stopsig;
+  VLOG(3) << "Stop sig: " << std::dec << stopsig;
```
JakeHillion, line 983: not sure if this was for debugging, but shouldn't this remain a 4?

JakeHillion (approved):
🤷‍♂️

PR #263, opened 2023-07-25
circleci: enable iwyu

JakeHillion:
## Summary

Enables iwyu on our code, ensuring we include what we use everywhere. Also picks up things like being able to forward declare a struct instead of include the header.

## Test plan

- CI


JakeHillion:
Pushed this to the main repo `iwyu` branch in case anyone wants to take another crack at it. Closing this PR as I don't intend to land it as is - the behaviour was weird and seemingly wrong.

PR #264, opened 2023-07-26
TypeGraph: Change Type::name() to return a reference instead of a copy

ajor:
Names which were generated on-demand are now stored in member variables, which are set during the ctor and can be regenerated when required (by NameGen).

### DIFF
--- a/oi/type_graph/Types.cpp
+++ b/oi/type_graph/Types.cpp
@@ -29,8 +29,8 @@ namespace oi::detail::type_graph {
 OI_TYPE_LIST
 #undef X
 
-std::string Primitive::name() const {
-  switch (kind_) {
+std::string Primitive::getName(Kind kind) {
+  switch (kind) {
     case Kind::Int8:
       return "int8_t";
     case Kind::Int16:
--- a/oi/type_graph/Types.h
+++ b/oi/type_graph/Types.h
@@ -80,8 +80,7 @@ class Type {
   virtual void accept(Visitor& v) = 0;
   virtual void accept(ConstVisitor& v) const = 0;
 
-  // TODO don't always return a copy for name()
-  virtual std::string name() const = 0;
+  virtual const std::string& name() const = 0;
   virtual size_t size() const = 0;
   virtual uint64_t align() const = 0;
   virtual NodeId id() const = 0;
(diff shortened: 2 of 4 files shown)


oi/type_graph/Types.cpp
```diff
@@ -29,8 +29,8 @@ namespace type_graph {
 OI_TYPE_LIST
 #undef X
 
-std::string Primitive::name() const {
-  switch (kind_) {
+std::string Primitive::getName(Kind kind) {
```
JakeHillion, line 32: If the overarching `name()` call returns `std::string_view` this shouldn't need to be specialised or ever stored on the heap.

oi/type_graph/Types.h
```diff
@@ -361,8 +360,10 @@ class Array : public Type {
 
   DECLARE_ACCEPT
 
-  virtual std::string name() const override {
-    return "OIArray<" + elementType_.name() + ", " + std::to_string(len_) + ">";
+  virtual const std::string& name() const override {
+    name_ = std::string{"OIArray<"} + elementType_.name() + ", " +
+            std::to_string(len_) + ">";
+    return name_;
```
JakeHillion, line 366: The ownership model and this field being mutable seems really suspicious. In a single threaded context, where constness makes much less sense anyway, this returned value is valid until the next call to `name()` - this is probably fine. In a multithreaded context, where mutable still applies, it's guaranteed to be valid never.

It seems like the copies are a lot safer to me. I looked at doing something clever with atomics to make this more explicit, but we'd still have no control over the lifetime of the return value. An alternative would be a reference counted string in an atomic that is only changed when necessary, which would guarantee the lifetime of the output with minimal heap movement.

Even putting a lock on this name wouldn't help much because you'd need to hold a read lock until the caller has finished using the result. I think the invariants of using name need to be documented if not codified, so at least if/when it bites us in future it won't be a complete surprise.

ajor: I'll admit this is kind of hacky, and this approach mainly came about through laziness...

Maybe a better solution would be to have a `setName()` function which can be called once by NameGen, after which `name()` can just return a reference to a `const std::string`

JakeHillion (approved):

oi/type_graph/Types.cpp
ajor: I never got around to deciding whether I thought it was better to return a `std::string` or `std::string_view` so I'm just going to merge as-in for now, but this is something that we could explore in the future.

PR #267, opened 2023-07-26
disciplined includes

JakeHillion:
## Summary

I got Include What You Use (IWYU) working on our repo in CI. It still has too many positives (both true and false) left to enable, so this is a (large) first round of fixes. IWYU ensures that every source file includes exactly what it uses, nothing more and nothing less. This gives you more freedom to move things around/forward declare things in other files as you know nothing is transitively including things.

## Test plan

- `make clobber; make configure-devel && make test`
- CI
- Will import this PR and check the internal build.

### DIFF
--- a/oi/type_graph/AddChildren.h
+++ b/oi/type_graph/AddChildren.h
@@ -16,11 +16,12 @@
 #pragma once
 
 #include <functional>
+#include <string>
+#include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
 #include "PassManager.h"
-#include "Types.h"
 #include "Visitor.h"
 
 class SymbolService;
--- a/oi/type_graph/AddPadding.cpp
+++ b/oi/type_graph/AddPadding.cpp
@@ -15,9 +15,15 @@
  */
 #include "AddPadding.h"
 
+#include <bitset>
 #include <cassert>
+#include <cstddef>
+#include <functional>
+#include <memory>
+#include <utility>
 
 #include "Flattener.h"
+#include "PassManager.h"
 #include "TypeGraph.h"
 
 template <typename T>
--- a/oi/type_graph/NameGen.cpp
+++ b/oi/type_graph/NameGen.cpp
@@ -15,6 +15,12 @@
  */
 #include "NameGen.h"
 
+#include <algorithm>
+#include <cstddef>
+#include <memory>
+#include <optional>
+
+#include "PassManager.h"
 #include "TypeGraph.h"
 
 template <typename T>
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -238,7 +238,8 @@ set(CMAKE_BUILD_RPATH
 )
 set(CMAKE_INSTALL_RPATH_USE_LINK_PATH TRUE)
 
-include_directories(SYSTEM "${DRGN_PATH}")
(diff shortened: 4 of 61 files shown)


tyroguru (approved):

oi/type_graph/AddPadding.cpp
```diff
@@ -15,9 +15,15 @@
  */
 #include "AddPadding.h"
 
+#include <bitset>
```
ajor, line 18: This one isn't used directly in this file at all - only through oi/Features.h

oi/type_graph/NameGen.cpp
```diff
@@ -15,6 +15,12 @@
  */
 #include "NameGen.h"
 
+#include <algorithm>
+#include <cstddef>
+#include <memory>
+#include <optional>
```
ajor, line 21: I also can't see how a few of these headers are needed for this class

ajor (commented):
There still seem to be a fairly large number of false positives and negatives included in this change

oi/type_graph/AddPadding.cpp
JakeHillion: Surprised me too, the reasoning it gave is it needs the implementation of `BitsetType::reference` for the result from `EnumBitset::operator[](T v)`. It's possible to make this one go away by adding `// IWYU pragma: export` to the `EnumBitset.h` header, which seems like a good idea.

oi/type_graph/NameGen.cpp
JakeHillion: Optional is used when you dereference `param.value`:
https://github.com/facebookexperimental/object-introspection/blob/55d1b1300960e4d096f7df7fdfe32977cfa1ecbc/oi/type_graph/NameGen.cpp#L120

The `<memory>` one I'm a bit stumped by. It wants it in loads of places because "// for allocator_traits<>::value_type". I think it includes everything you need for a templated function's implementation, so perhaps string is using it.

oi/type_graph/AddChildren.h
```diff
@@ -16,11 +16,12 @@
 #pragma once
 
 #include <functional>
+#include <string>
+#include <unordered_map>
 #include <unordered_set>
 #include <vector>
 
 #include "PassManager.h"
-#include "Types.h"
```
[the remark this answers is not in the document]
JakeHillion: This one is weird/wrong. In several places it tried to replace "Types.h" with "PassManager.h". I think it's confused by the forward declarations in "Types.h" and assumes they're not fully defined... I'll put the removed `Types.h` back for this change.

ajor:
Thanks for the explanations. I think the disconnect for me was that I was considering something like the return value of a function as an implementation detail, where the header would be free to change the exact type as long as it was compatible with the code that was using it.

e.g. For a function:
```
header:
  std::optional<int> foo();

consumer:
  auto x = *foo();
```
The return type could be changed to something else without needing any changes in the caller:
```
header:
  std::unique_ptr<int> foo();

consumer:
  auto x = *foo();
```

So I wouldn't have personally included the `optional` or `memory` headers in either case, unless the caller is explicit about the type, e.g.:
```
std::optional<int> f = foo();
auto x = *f;
```

Maybe I'm just wrong about this though...

JakeHillion:
Might take another approach to this at some point in the future, but dropping it in its current form.

PR #275, opened 2023-08-02
Printer: Use NodeIds for stable IDs when printing graphs

ajor:
This removes Printer's legacy behaviour of generating an ID for each node as it gets printed. This old method meant that if new nodes were added to or removed from a graph, every ID after the new/removed node would change.

Now IDs are stable so it is easier to follow specific nodes through multiple transformation passes in CodeGen.

### DIFF
--- a/oi/type_graph/TypeGraph.h
+++ b/oi/type_graph/TypeGraph.h
@@ -48,7 +48,7 @@ class TypeGraph {
     rootTypes_.push_back(type);
   }
 
-  NodeTracker& resetTracker() noexcept;
+  NodeTracker& resetTracker() const noexcept;
 
   // Override of the generic makeType function that returns singleton Primitive
   // objects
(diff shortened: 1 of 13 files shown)


oi/type_graph/TypeGraph.h
```diff
@@ -87,7 +88,7 @@ class TypeGraph {
   std::vector<std::reference_wrapper<Type>> rootTypes_;
   // Store all type objects in vectors for ownership. Order is not significant.
   std::vector<std::unique_ptr<Type>> types_;
-  NodeTracker tracker_;
+  mutable NodeTracker tracker_;
```
ajor, line 91: This is a bit of a hack, but I remove NodeTracker from TypeGraph entirely in the next PR: #284

JakeHillion (approved):

PR #281, opened 2023-08-10
codegen: split genClassTypeHandler

JakeHillion:
## Summary

This function is pretty long and does three distinct things: creates the structure for a class type handler, figures out the appropriate static type, and generates the traversal function. Split it into distinct functions. This is motivated by needing to add additional logic for OIL/TreeBuilder V2 and the function becoming unwieldy.

Cleaned up indentation and such in the generated code while I was in there.

## Test plan

CI


oi/CodeGen.cpp
```diff
@@ -597,143 +597,167 @@ void addStandardTypeHandlers(std::string& code) {
...
+void CodeGen::genClassTypeHandler(const Class& c, std::string& code) {
+  std::string helpers;
 
-      if (thriftIssetMember != nullptr && thriftIssetMember != &member) {
-        traverser += "\n  .write(getThriftIsset(t, " + std::to_string(i) + "))";
-      }
+  const Member* thriftIssetMember = nullptr;
+  if (const auto it = thriftIssetMembers_.find(&c);
+      it != thriftIssetMembers_.end()) {
+    thriftIssetMember = it->second;
```
ajor, line 732:
```suggestion
    const Member& thriftIssetMember = *it->second;
```
Minor, but a little simplification is possible now compared to the old code - no need to keep a pointer out of the scope of the if-block

ajor (approved):

PR #282, opened 2023-08-11
type_graph: provide inputName for all types

JakeHillion:
## Summary

Type Graph deduplicates and modifies names to better fit the generated code, for example `int32_t[4]` becomes `OIArray<int32_t, 4>` and `struct MyStruct` might become `struct MyStruct_0`.

Add an `inputName` which better represents the original input code which can be used when building the tree.

## Test plan

Updated `test_codegen.cpp` and `test_name_gen.cpp` to respect these changes. This is more to assert any changes are deliberate rather than to claim the current ones are correct...

- CI

### DIFF
--- a/test/test_codegen.cpp
+++ b/test/test_codegen.cpp
@@ -63,7 +63,7 @@ TEST(CodeGenTest, TransformContainerAllocator) {
       Param
         Primitive: int32_t
       Param
-        DummyAllocator (size: 8)
+        DummyAllocator [MyAlloc] (size: 8)
           Primitive: int32_t
 )");
 }
@@ -98,7 +98,7 @@ TEST(CodeGenTest, TransformContainerAllocatorParamInParent) {
       Param
         Primitive: int32_t
       Param
-        DummyAllocator (size: 0)
+        DummyAllocator [MyAlloc<std::pair<const int, int>>] (size: 0)
 [3]       Container: std::pair<int32_t const, int32_t> (size: 8)
             Param
               Primitive: int32_t
@@ -121,16 +121,16 @@ TEST(CodeGenTest, RemovedMemberAlignment) {
         Primitive: int8_t
 )",
                 R"(
-[0] Class: MyClass_0 (size: 24, align: 8)
-      Member: a_0 (offset: 0, align: 1)
+[0] Class: MyClass_0 [MyClass] (size: 24, align: 8)
+      Member: a_0 [a] (offset: 0, align: 1)
         Primitive: int8_t
       Member: __oi_padding_1 (offset: 1)
-[1]     Array: (length: 15)
+[1]     Array: [int8_t[15]] (length: 15)
           Primitive: int8_t
-      Member: c_2 (offset: 16, align: 1)
+      Member: c_2 [c] (offset: 16, align: 1)
         Primitive: int8_t
       Member: __oi_padding_3 (offset: 17)
-[2]     Array: (length: 7)
+[2]     Array: [int8_t[7]] (length: 7)
           Primitive: int8_t
 )");
 }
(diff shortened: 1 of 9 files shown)


test/test_codegen.cpp
```diff
@@ -177,16 +177,16 @@ TEST(CodeGenTest, RemovedMemberAlignment) {
         Primitive: int8_t
 )",
                 R"(
-[0] Class: MyClass_0 (size: 24, align: 8)
-      Member: a_0 (offset: 0, align: 1)
+[0] Class: MyClass_0 [MyClass] (size: 24, align: 8)
+      Member: a_0 [a] (offset: 0, align: 1)
```
JakeHillion, line 181: This member output from Printer seems particularly unnecessary - it'll only ever be removing the `_{i}` or leaving the dots in vptrs. I erred on the side of printing all of the information in the type graph though.

ajor (commented):
If we have two vectors:
```
std::vector<int, std::allocator<int>>
std::vector<int, MyLargeAllocator<int>>
```

What would the input names of these be calculated as? They'd need to show up differently for users to understand why one vector has a larger static size than the other.

I suspect that we'll need to keep input names for dummy types.

JakeHillion:
> If we have two vectors:
> 
> ```
> std::vector<int, std::allocator<int>>
> std::vector<int, MyLargeAllocator<int>>
> ```
> 
> What would the input names of these be calculated as? They'd need to show up differently for users to understand why one vector has a larger static size than the other.
> 
> I suspect that we'll need to keep input names for dummy types.

From `OilIntegration.std_vector_del_allocator_a`:

```
[
{
  "name" : "a0",
  "typePath" : ["a0"],
  "typeNames" : ["Foo"],
  "staticSize": 80,
  "dynamicSize": 0,
  "exclusiveSize": 0,
  "members": [
    {
      "name" : "v1",
      "typePath" : ["a0", "v1"],
      "typeNames" : ["std::vector<C, >"],
      "staticSize": 40,
      "dynamicSize": 0,
      "exclusiveSize": 40,
      "pointer": 140727394660736,
      "length": 1,
      "capacity": 1,
      "members": []
    },
    {
      "name" : "v2",
      "typePath" : ["a0", "v2"],
      "typeNames" : ["std::vector<C, >"],
      "staticSize": 40,
      "dynamicSize": 0,
      "exclusiveSize": 40,
      "pointer": 140727394660776,
      "length": 2,
      "capacity": 2,
      "members": []
    }
    ]
}
]
```

Spot on. This is pretty useless. I'll figure out the input names and plumb them through.

JakeHillion:
Much better now:
```
[
{
  "name" : "a0",
  "typePath" : ["a0"],
  "typeNames" : ["Foo"],
  "staticSize": 48,
  "dynamicSize": 0,
  "exclusiveSize": 0,
  "members": [
    {
      "name" : "v1",
      "typePath" : ["a0", "v1"],
      "typeNames" : ["std::vector<C, CustomAllocator<ns_std_vector_del_allocator::nsA::C>>"],
      "staticSize": 24,
      "dynamicSize": 0,
      "exclusiveSize": 24,
      "pointer": 140722823798016,
      "length": 1,
      "capacity": 1,
      "members": []
    },
    {
      "name" : "v2",
      "typePath" : ["a0", "v2"],
      "typeNames" : ["std::vector<C, CustomAllocator<ns_std_vector_del_allocator::nsB::C>>"],
      "staticSize": 24,
      "dynamicSize": 0,
      "exclusiveSize": 24,
      "pointer": 140722823798040,
      "length": 2,
      "capacity": 2,
      "members": []
    }
    ]
}
]
```

Some of the unit test output makes it look like this doesn't work, but that's because the Printer gets the inputName from what you give it instead of the DWARF. When the input name is generated from `DrgnParser` it doesn't have the template parameters stripped so looks correct.

I did debate regenerating the input name from the graph (e.g. child template params inputName()) but for now trusting the DWARF is easier and we can change it if it becomes a problem.

ajor (approved):
Looks good!

PR #283, opened 2023-08-11
DrgnParser: Don't error out for certain bad DWARF

ajor:
Match the behaviour of CodeGen v1 and just log warnings when certain debug info has issues.

Example warning:
```
W0811 10:20:40.967265 112689 DrgnParser.cpp:522] Error looking up parent (1) for type 'EventBase': 2 DW_TAG_member has invalid DW_AT_data_member_location
```

### DIFF
--- a/oi/type_graph/DrgnParser.cpp
+++ b/oi/type_graph/DrgnParser.cpp
@@ -71,6 +71,10 @@ Primitive::Kind primitiveFloatKind(struct drgn_type* type) {
   }
 }
 
+void warnForDrgnError(struct drgn_type* type,
+                      const std::string& msg,
+                      struct drgn_error* err);
+
 }  // namespace
 
 Type& DrgnParser::parse(struct drgn_type* root) {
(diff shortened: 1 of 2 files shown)


oi/type_graph/DrgnParser.cpp
```diff
@@ -71,6 +71,10 @@ Primitive::Kind primitiveFloatKind(struct drgn_type* type) {
   }
 }
 
+void warnForDrgnError(struct drgn_type* type,
```
JakeHillion, line 74: It would need passing through but could we consider the strict mode I added recently here? If in strict mode throw, else warn.

ajor: I'm not sure about this - it's down to bad DWARF so there's no action that users or OI developers could take to resolve this warning/error. That means they'd have to disable strict mode in order to progress, but now that strict mode's disabled they might miss any other errors that appear later.

PR #285, opened 2023-08-15
OIL v2

JakeHillion:
## Summary

Updates Object Introspection as a Library (OIL) to produce the full output tree of an introspected type, similar to OID.

## Test plan

- CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -334,9 +334,18 @@ target_link_libraries(treebuilder
 
 ## OI Outputs
 ### Object Introspection as a Library (OIL)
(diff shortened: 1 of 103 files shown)


.circleci/config.yml
```diff
@@ -208,7 +208,7 @@ jobs:
           name: Test
           environment:
             # disable drgn multithreading as tests are already run in parallel
-            OMP_NUM_THREADS: 1
+            OMP_NUM_THREADS: 2
```
JakeHillion, line 211: It would be really great if this was resolved before landing.

PR #288, opened 2023-08-17
TypeGraph: Add option to record enumerator names

ajor:
It's not enabled yet, but should be useful for the key capture work.


JakeHillion (approved):
Nice! Would we turn this on whenever any key capture is provided in the config? It seems quite hard to do based on whether there is a relevant key to capture due to the ordering.

ajor:
Yeah exactly, I was just planning on enabling this whenever there's any key-capture requested in the config

Issue #303, opened 2023-08-18
std::unordered_map is not implemented for treebuilder v2


JakeHillion:
Examples:
- `OilIntegration.std_unordered_map_custom_operator_a`

PR #329, opened 2023-08-25
Implement Container V2 for `unordered` containers

### LINKED ISSUE #302: std::unordered_set is not implemented for treebuilder v2
Examples:
- `OilIntegration.std_unordered_set_custom_operator_a`

ttreyer:
## Summary
Implement Container V2 for each `unordered` containers.
The size overhead was measured using: https://godbolt.org/z/b4qrr3YdM
Closes #302 and #303.

## Test plan
```
$ build/test/integration/integration_test_runner --gtest_filter='OilIntegration.std_unordered_set_custom_operator_a' --verbose
```

### DIFF
--- a/types/unordered_set_type.toml
+++ b/types/unordered_set_type.toml
@@ -67,3 +67,71 @@ struct TypeHandler<DB, %1%<T0, T1, T2, T3>> {
   }
 };
 """
+
+traversal_func = """
+auto tail = returnArg
+  .write((uintptr_t)&container)
+  .write(container.bucket_count())
+  .write(container.size());
+
+for (const auto &it : container) {
+  tail = tail.delegate([&it](auto ret) {
+    return OIInternal::getSizeType<DB>(it, ret);
+  });
+}
+
+return tail.finish();
+"""
+
+[[codegen.processor]]
+type = "types::st::VarInt<DB>"
+func = """
+el.pointer = std::get<ParsedData::VarInt>(d.val).value;
+"""
+
+[[codegen.processor]]
+type = "types::st::VarInt<DB>"
+func = """
+// Using the container's capacity to temporarily store the number of buckets
+// TODO: Is there another way to pass a value across processors?
+el.container_stats.emplace(result::Element::ContainerStats {
+  .capacity = std::get<ParsedData::VarInt>(d.val).value,
+});
+"""
+
+[[codegen.processor]]
+type = "types::st::List<DB, typename TypeHandler<DB, T0>::type>"
+func = """
+#ifdef __GLIBCXX__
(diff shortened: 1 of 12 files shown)


types/unordered_set_type.toml
```diff
@@ -67,3 +67,78 @@ struct TypeHandler<DB, %1%<T0, T1, T2, T3>> {
...
+type = "types::st::VarInt<DB>"
+func = """
+el.pointer = std::get<ParsedData::VarInt>(d.val).value;
+"""
+
+[[codegen.processor]]
+type = "types::st::VarInt<DB>"
+func = """
+// Using the container's capacity to temporarily store the number of buckets
+// TODO: Is there another way to pass a value across processors?
```
JakeHillion, line 96: I think this is a smart use of what's there for now. The current alternative is to make the next processor take a pair, but as discussed it's messy in the current implementation.
Future options:
- Make the interface for taking out of a complex type in a processor better (my preferred).
- Add an option for naming variables like `VarInt` and making them available in future processors (complex to add to the current implementation of separate functions).
- Having some sort of dump space where variables can be carried through (complex to manage).

types/unordered_set_type.toml
```diff
@@ -67,3 +67,78 @@ struct TypeHandler<DB, %1%<T0, T1, T2, T3>> {
...
+static constexpr size_t bucket_size = sizeof(OI_bucket);
+#else
+static_assert(false && "No known element_size for sets. See types/set_type.toml");
+#endif
+
+static constexpr std::array<std::string_view, 1> names{"TODO"};
+static constexpr auto childField = inst::Field{
+  sizeof(T0),
+  "[]",
+  names,
```
JakeHillion, line 126: This is fixed now. You can use `NameProvider<T0>::names`. Alternatively you can land after #330 and replace this with a `make_field<DB, T0>("[]")` call.

types/unordered_set_type.toml
```diff
@@ -67,3 +67,78 @@ struct TypeHandler<DB, %1%<T0, T1, T2, T3>> {
...
+  TypeHandler<DB, T0>::fields,
+  TypeHandler<DB, T0>::processors,
+};
+
+auto list = std::get<ParsedData::List>(d.val);
+// Reading the bucket count that was stored in `capacity` by the processor above.
+el.exclusive_size += el.container_stats->capacity * bucket_size;
```
JakeHillion, line 133: nit: I would name a variable `bucket_count` here, just for clarity.

JakeHillion (approved):
These look fantastic! Thanks for taking the time to give us great container implementations.

PR #332, opened 2023-08-29
Compile Time OIL for OIL v2

JakeHillion:
Summary:
Update `OIGenerator` and out BUCK stuff for compile time OIL with OIL v2. Main things:
- Switch `OIGenerator` from the `getObjectSize` call to the new `introspect` call.
- Switch from looking at template parameters to looking at function parameters, as this was exposing a bug in our elfutils/drgn and this way it's the same as OID.
- Migrate `OIGenerator` to CodeGen v2 and update CodeGen v2 to accept a linkage name.
- Update the compile time example to be the same as the JIT example, using the new interface and the JSON exporter.
- Clean up the `ObjectIntrospection.h` header.

Differential Revision: D48687728

fbshipit-source-id: 2c3c041fd1b6499c5e02eb5e2082a977bfa529d7


oi/FuncGen.cpp
```diff
@@ -274,6 +274,28 @@ void __attribute__((used, retain)) introspect_%2$016x(
...
+                                            const std::string& type,
+                                            const std::string& linkageName) {
+  std::string typeHash =
+      (boost::format("%1$016x") % std::hash<std::string>{}(type)).str();
+
+  code += "/* RawType: ";
+  code += type;
+  code += " */\n";
+  code += "extern \"C\" IntrospectionResult ";
+  code += linkageName;
```
JakeHillion, line 287: This thing throws a warning because it's a complex C++ type being returned with C style linkage. However it seems to work fine, so I'm inclined not to care. The alternative could be to use a demangled name (which we get by demangling the mangled name) and hope it gets mangled the same again, but this seems more flaky.

tyroguru (approved):

PR #336, opened 2023-09-04
Tweak social media previews for Oid website

jgkamat:
Hello, hope everyone is doing well! :)

## Summary
I noticed that the website has some default text in the social media preview. I went ahead and tweaked it to not be the default anymore.

In addition, it seems like the favicon dosen't seem to work: https://realfavicongenerator.net/favicon_checker?protocol=https&site=facebookexperimental.github.io%2Fobject-introspection%2F

I attempted to fix this by making the link relative.

## Test plan
Serve website via docusaurus instructions and verify title/meta tag is updated. Also verify that favicon shows up. 

To "test" favicon in production, I removed the leading `/` from the favicon on the live site using inspect element, after which I could see the favicon again. I'll try to follow up to make sure things are working properly on the site for this one.

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -48,8 +48,8 @@ const config = {
   headTags: [
     // Favicons declarations
     { tagName: 'link', attributes: { rel: 'apple-touch-icon', sizes: '180x180', href: '/img/apple-touch-icon.png' } },
-    { tagName: 'link', attributes: { rel: 'icon', sizes: '32x32', href: '/img/favicon-32x32.png' } },
-    { tagName: 'link', attributes: { rel: 'icon', sizes: '16x16', href: '/img/favicon-16x16.png' } },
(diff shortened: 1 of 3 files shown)


website/docusaurus.config.js
```diff
@@ -48,8 +48,8 @@ const config = {
   headTags: [
     // Favicons declarations
     { tagName: 'link', attributes: { rel: 'apple-touch-icon', sizes: '180x180', href: '/img/apple-touch-icon.png' } },
```
jgkamat, line 50: If this ends up working, I think these other links might need to be made relative too. I remember there being some nuance to this though, so hopefully this is the right change...

ajor: Let's see how it goes 👍

ajor (approved):
Hello Jay and thanks for catching this!

PR #338, opened 2023-09-06
Implement Container V2 for F14Map

ttreyer:
## Summary
Implement Container V2 for `folly::F14VectorMap`.

## Test plan
Tests in `folly_f14_vector_map.toml` cover this implementation, but we'll have to wait for the CI to confirm they pass.

### DIFF
--- a/types/f14_node_map.toml
+++ b/types/f14_node_map.toml
@@ -71,6 +71,9 @@ struct TypeHandler<DB, %1%<T0, T1, T2, T3, T4>> {
 """
 
 traversal_func = """
+// TODO: This implementation enables the traversal of the container,
+//       but doesn't report the memory footprint accurately.
+//       Revisit this implementation and fix memory footprint reporting.
 auto tail = returnArg
   .write((uintptr_t)&container)
   .write(container.size());
--- a/types/f14_node_map.toml
+++ b/types/f14_node_map.toml
@@ -35,3 +35,87 @@ void getSizeType(const %1%<Key, Mapped, Hasher, KeyEqual, Alloc> &container, siz
     }
 }
 """
+
+handler = """
+template <typename DB, typename T0, typename T1, typename T2, typename T3, typename T4>
+struct TypeHandler<DB, %1%<T0, T1, T2, T3, T4>> {
+  using type = types::st::Pair<DB,
+    types::st::VarInt<DB>,
+    types::st::Pair<DB,
+      types::st::VarInt<DB>,
+      types::st::List<DB, types::st::Pair<DB,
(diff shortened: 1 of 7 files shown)


ttreyer:
I'm not sure my additions to `oi/OITraceCode.cpp` are the right way to do this. I'd love some feedback on that particular part.

ajor:
> I'm not sure my additions to `oi/OITraceCode.cpp` are the right way to do this. I'd love some feedback on that particular part.

There are a few strange things I see going on in the old f14_vector_map.toml file:
- `numTemplateParams` is specified to be 2 and `replaceTemplateParamIndex` is not defined, which means the HeterogeneousAccessHash template parameter is not being stubbed out
- `allocatorIndex` is not specified, so the allocator parameter is not being removed

I'm not sure how this has been working previously, but without looking into and fixing those other problems, I think a better workaround for the HeterogeneousAccessHash issue would be to force CodeGen v1 to include the folly namespace with:
```
ns = ["namespace folly"]
```

All of these issues are for CodeGen v1 only, so I wouldn't spend much time investigating them as v1 will hopefully be going away entirely soon.

ttreyer:
I kept the "namespace folly" out for now and stubbed both parameters out. Both F14NodeMap and F14FastMap had proper values for `replaceTemplateParamIndex` and `allocatorIndex`, but only F14VectorMap was missing these. Since we don't need these template parameters to traverse the containers, I'm happy to go that way!

types/f14_node_map.toml
```diff
@@ -35,3 +35,87 @@ void getSizeType(const %1%<Key, Mapped, Hasher, KeyEqual, Alloc> &container, siz
...
+        typename TypeHandler<DB, T1>::type>>>>;
+
+  static types::st::Unit<DB> getSizeType(
+      const %1%<T0, T1, T2, T3, T4>& container,
+      typename TypeHandler<DB, %1%<T0, T1, T2, T3, T4>>::type returnArg) {
+    size_t memorySize = container.getAllocatedMemorySize();
+    auto tail = returnArg
+      .write(memorySize)
+      .write(container.bucket_count())
+      .write(container.size());
```
ajor, line 57: These look like the right things to save, but `traversal_func` isn't including them.

Side note: is this `handler` used for anything now the new `traversal_func` method exists?

ajor (changes requested):
I think the json_v2 results for these types are wrong: they're all showing sizes of 24 no matter how many elements are in the containers and don't match the v1 results.

ajor (approved):
Fair enough, yeah let's merge then but not forget!

PR #357, opened 2023-09-21
prevent name collision between uintptr_t and uint64_t

JakeHillion:
## Summary

`uintptr_t` is a typedef of `uint64_t` but doesn't appear as such in the type graph. This meant if we generated a `uintptr_t` and also had a `uint64_t` in the output, they would get a `NameProvider<T>` each and this would cause a compilation failure.

Stop generating `uintptr_t` for these cases where we're stubbing pointers (the only cases it gets generated in). Change this to a special enum class called `StubbedPointer` with the same memory footprint. This will make stubbed pointers easier to search for in the output and prevents the type collision of the primitives.

## Test plan

- CI
- `make test`

### DIFF
--- a/oi/OITraceCode.cpp
+++ b/oi/OITraceCode.cpp
@@ -143,3 +143,5 @@ struct validate_offset {
   static constexpr bool value = true;
   static_assert(ExpectedOffset == ActualOffset);
 };
+
+enum class StubbedPointer : uintptr_t {};
--- a/oi/type_graph/DrgnParser.cpp
+++ b/oi/type_graph/DrgnParser.cpp
@@ -464,13 +464,13 @@ static drgn_type* getPtrUnderlyingType(drgn_type* type) {
 
 Type& DrgnParser::enumeratePointer(struct drgn_type* type) {
   if (!chasePointer()) {
-    return makeType<Primitive>(type, Primitive::Kind::UIntPtr);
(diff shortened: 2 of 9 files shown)


ajor (commented):
Would it work if we didn't add the specialisation of `uint64_t`, and only skipped over the standard NameProvider generation for `uintptr_t`?

`uintptr_t` is always an alias, but it might not always be an alias for `uint64_t`. It'd seem cleaner to only special case the alias (`uintptr_t`) and not regular primitive types, and would make it work no matter what the underlying type is.

ajor (commented):
The code looks great, I just have a concern about potential name collisions with our `StubbedPointer` and types defined in the target application.

There wouldn't be a collision in the generated code due to NameGen appending numbers to user types, but in the output (e.g. the JSON in the test) both our type and a user's type would show as "StubbedPointer", right?

We could reduce the chance of collision with a name like "OIStubbedPointer", but I think to get rid of it entirely we'd need to output a name that would be invalid as a type name, e.g. "uintptr_t (stubbed)" - no user defined type could have spaces or brackets in its name.

JakeHillion:
This new `inputName()` value isn't output by printer, but I don't think that really matters - you can infer the value from the value of `name()` that is output.

ajor (approved):

PR #358, opened 2023-09-21
Use objectintrospection.org for the website

zpao:
## Summary
This updates the docusaurus config and adds the CNAME file for gh pages, then updates the readme to point at this domain.

## Test plan
Build website & serve it to ensure it works from the root.
```sh
cd website
yarn
yarn build
yarn serve
```

### DIFF
--- a/website/docusaurus.config.js
+++ b/website/docusaurus.config.js
@@ -13,8 +13,8 @@
 const config = {
   title: 'Object Introspection',
   tagline: 'Dynamic C++ Object Profiling',
-  url: 'https://facebookexperimental.github.io',
-  baseUrl: '/object-introspection',
+  url: 'https://objectintrospection.org',
+  baseUrl: '/',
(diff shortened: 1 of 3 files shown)


website/docusaurus.config.js
```diff
@@ -24,7 +24,7 @@ const config = {
   // If you aren't using GitHub pages, you don't need these.
   organizationName: 'facebookexperimental', // Usually your GitHub org/user name.
   projectName: 'object-introspection', // Usually your repo name.
-	deploymentBranch: 'web',
+	deploymentBranch: 'gh-pages',
```
zpao, line 27: Note that this isn't actually important since you aren't using docusaurus directly to deploy (`yarn deploy`) but good to be correct.

JakeHillion (approved):
Really appreciate you helping us out with this Paul!

PR #364, opened 2023-09-22
Implement Map Key Capture

ajor:
Big PR, but almost half of it is tests. I'd recommend looking through it commit-by-commit. 

This PR adds support for capturing map keys to OI. Maps to be captured are identified in the OI config section "capture_keys", with a similar format to the ignored members feature, i.e. we take a class name and member names and any capture keys for those members if they are containers.

Only `std::map` and `std::unordered_map` are supported in this PR, but supporting others will just involve updating their container TOMLs.

### Example
For a map:
```
std::map<int, int> mymap;
mymap[1] = 2;
mymap[3] = 4;
```
We would output the keys in the `data` field, as shown:
```
[{
  # mymap
  "members": [
    {
      # First map element
      "data": 1,
      "members": [
        {"name": "key", ...},
        {"name": "value", ...}
      ] 
    },
    {
      # Second map element
      "data": 3,
      "members": [
        {"name": "key", ...},
        {"name": "value", ...}
      ] 
    } 
  ] 
}]
```

See the new integration tests in `capture_keys.toml` for more examples.

### Not supported
- Converting enum underlying values into their user-facing enumerator names. They're just shown as ints for now.
- Propagating captured keys into the type path, or a similar list. Wouldn't be hard to do.
- Capturing keys from maps inside maps, i.e. only the `int` keys from the out map would be captured for `std::map<int, std::map<long, char>>`. As the inner map is a template parameter of the outer one, changing this inner map by marking it for key capture would also require us to change the allocator param of the outer map.


types/README.md
```diff
@@ -48,6 +48,10 @@ This document describes the format of the container definition files contained i
   `static types::st::Unit<DB> getSizeType(const T& container, ST returnArg)`
   where `ST` defines the combined static type of each supplied processor.
 
+- `extra`
+
+  Any extra C++ code to be included directly.
```
JakeHillion, line 53: nit: can this mention that there's no namespacing on this code? it might catch people unaware that they can't collide with any other container implementation.

include/oi/result/Element.h
```diff
@@ -40,6 +42,7 @@ struct Element {
   size_t exclusive_size;
 
   std::optional<uintptr_t> pointer;
+  std::variant<std::nullopt_t, uint64_t, std::string> data = {std::nullopt};
```
JakeHillion, line 45: It would make more sense to me to disambiguate the pointer/general number entry here. You can do:
```
struct Pointer { uintptr_t p; };
struct Scalar { uint64_t s; };
std::variant<std::nullopt_t, Pointer, Scalar, std::string> data;
```
Which leaves the interface still quite clean.

This would require changing the non-specialised capture function to a `Sum` though. Your choice.

oi/CodeGen.cpp
```diff
@@ -983,6 +1069,10 @@ void CodeGen::addTypeHandlers(const TypeGraph& typeGraph, std::string& code) {
     } else if (const auto* con = dynamic_cast<const Container*>(&t)) {
       genContainerTypeHandler(config_.features, definedContainers_,
                               con->containerInfo_, con->templateParams, code);
+    } else if (const auto* cap = dynamic_cast<const CaptureKeys*>(&t)) {
+      genContainerTypeHandler(config_.features, definedContainers_,
+                              cap->containerInfo(),
+                              cap->container().templateParams, code);
```
JakeHillion, line 1075: is this necessary? the topo sorter tests make it look like this will visit the container twice and get stopped by the set check

oi/ContainerInfo.h
```diff
@@ -71,12 +72,15 @@ struct ContainerInfo {
...
-  ContainerInfo& operator=(const ContainerInfo& other) = delete;
-
   ContainerInfo(ContainerInfo&&) = default;
   ContainerInfo& operator=(ContainerInfo&&) = default;
 
+  // Explicit interface for copying
+  ContainerInfo clone() const {
+    ContainerInfo copy{*this};
+    return copy;
+  }
```
JakeHillion, line 82: I love this! We should make all types everywhere like this. 🦀

JakeHillion (approved):

oi/CodeGen.cpp
ajor: The CaptureKeys node has a different ContainerInfo from the map that it contains. The ContainerInfo is copied and modified in the KeyCapture pass

PR #365, opened 2023-09-22
CodeGen: Store list of ContainerInfos in unique_ptrs for reference stability

ajor:
Lots of places rely on reference stability of ContainerInfo objects (CodeGen's deduplication, Container nodes' containerInfo_ member).

In the key capture work, we need to be able to append to this list, which would invalidate references before this change.

`std::deque<ContainerInfo>` would also be another option we could go with.


JakeHillion (approved):
If you want a fancier solution, I vote against deque and for making ContainerInfo a Type! More messing about and the storage would be the same though (vector of unique_ptr, and probably a vector of reference_wrapper for fast lookup).

LGTM.

ajor:
Maybe another time ;)

PR #366, opened 2023-09-26
jitlog: use a memfd and glog

JakeHillion:
jitlog: use a memfd and glog

### DIFF
--- a/oi/OIDebugger.cpp
+++ b/oi/OIDebugger.cpp
@@ -29,6 +29,7 @@
 #include <cerrno>
 #include <chrono>
 #include <cstddef>
+#include <cstdio>
 #include <cstring>
 #include <numeric>
 #include <span>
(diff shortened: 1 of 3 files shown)


JakeHillion:
It seems like if this runs on the same process twice there's an issue. Debugging now.

JakeHillion:
Now works fine when run twice. New issue is it can't close the memfd in the tracee when cleaning up due to a ptrace error. This might be because it's called after dropping ptrace? Will come back to this.

JakeHillion:
I've moved `cleanupLogFile` into `restoreState`. The process needs to be in a stopped state for `remoteSyscall`'s ptrace commands to work, and this seemed appropriate because it is part of putting the process back to how it was. This is in the hot path as it involves stopping the target, but given how infrequently JIT logging is used I'm not worried. It's likely possible to split `cleanupLogFile` into the remote syscall bit and the printing bit, but for now I'd rather keep it simple and optimise it in future if necessary.

By the looks of `restoreState` and the comments in `remoteSyscall` I think this location is fine. `remoteSyscall` looks pretty careful to reset registers to how they were on entry, and the process is definitely stopped at this point. Would appreciate @tyroguru's feedback on this before merging though.

tyroguru (approved):

PR #370, opened 2023-09-28
Add the `Incomplete` Type node

ttreyer:
## Summary
Any Incomplete type encountered by the DrgnReader gets effectively turned into a Primitive::Kind::Void. This makes debugging these issues fairly difficult, as we lose all information regarding the Incomplete type.

This PR changes this behaviour, so that we can still identify the type that got classified as Incomplete.

## Test plan
Update the test "DrgnParserTest.PointerIncomplete" to take the new `Incomplete` node into account.
Update the `TypeGraphParser` to rebuild the `Incomplete` node.

### DIFF
--- a/oi/type_graph/Types.h
+++ b/oi/type_graph/Types.h
@@ -34,6 +34,7 @@
 #include <optional>
 #include <string>
 #include <string_view>
+#include <variant>
 #include <vector>
 
 #include "oi/ContainerInfo.h"
@@ -186,6 +187,10 @@ class Incomplete : public Type {
   Incomplete(Type& underlyingType) : underlyingType_(underlyingType) {
   }
 
+  Incomplete(std::string underlyingTypeName)
+      : underlyingType_(std::move(underlyingTypeName)) {
+  }
+
   static inline constexpr bool has_node_id = false;
 
   DECLARE_ACCEPT
@@ -195,7 +200,13 @@ class Incomplete : public Type {
   }
 
   std::string_view inputName() const override {
-    return underlyingType_.inputName();
+    if (std::holds_alternative<std::string>(underlyingType_)) {
+      return std::get<std::string>(underlyingType_);
+    }
+
+    return std::get<std::reference_wrapper<Type>>(underlyingType_)
+        .get()
+        .inputName();
   }
 
   size_t size() const override {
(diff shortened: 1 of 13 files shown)


oi/type_graph/Types.h
```diff
@@ -175,6 +176,48 @@ struct TemplateParam {
...
+ public:
+  Incomplete(Type& underlyingType) : underlyingType_(underlyingType) {
+  }
+
+  static inline constexpr bool has_node_id = false;
+
+  DECLARE_ACCEPT
+
+  const std::string& name() const override {
+    return underlyingType_.name();
```
JakeHillion, line 194: I think this should be `void` to generate code that would work as well as it did before.

oi/type_graph/Types.h
```diff
@@ -175,6 +177,63 @@ struct TemplateParam {
...
+  }
+
+  std::string_view inputName() const override {
+    if (std::holds_alternative<std::string>(underlyingType_)) {
+      return std::get<std::string>(underlyingType_);
+    }
+
+    return std::get<std::reference_wrapper<Type>>(underlyingType_)
+        .get()
+        .inputName();
```
JakeHillion, line 209:
```suggestion
    return std::visit([](const auto& d) -> std::string_view {
      using T = std::decay_t<decltype(d)>;
      if constexpr (std::is_same_v<T, std::string>) {
        return d;
      } else {
        return d.get().inputName();
      }
    }, underlyingType_);
```

No joy?

oi/type_graph/Types.h
```diff
@@ -175,6 +177,63 @@ struct TemplateParam {
...
+  NodeId id() const override {
+    return -1;
+  }
+
+  std::optional<std::reference_wrapper<Type>> underlyingType() const {
+    if (std::holds_alternative<std::string>(underlyingType_)) {
+      return std::nullopt;
+    }
+
+    return std::get<std::reference_wrapper<Type>>(underlyingType_);
```
JakeHillion, line 229:
```suggestion
    return std::visit([](const auto& d) -> std::optional<std::reference_wrapper<Type>> {
      using T = std::decay_t<decltype(d)>;
      if constexpr (std::is_same_v<T, std::string>) {
        return std::nullopt;
      } else {
        return d;
      }
    }, underlyingType_);
```

JakeHillion (approved):
It would be better to use `std::visit`, but otherwise this change looks great! I like how it simplifies some of the primitive logic too.

JakeHillion:
Merging this as is because it's amazing for debugging these issues. Thanks for this!

PR #371, opened 2023-09-28
support 0 to many config files

JakeHillion:
## Summary

Previously OID/OIL required exactly one configuration file. This change makes it so you can supply 0 or more configuration files. 0 is useful if you have pre-generated the cache or use some sort of remote generation system. 1 is useful for the common case, where you have a configuration file that describes your entire source and use just that. More are useful if you have supplemental bits of config you wish to apply/override - see the changes to the integration test framework where we do exactly this.

## Test plan

This isn't super well tested. It works for the test cases which add features via the config or enable `codegen.ignore`.

- CI

### DIFF
--- a/oi/Config.cpp
+++ b/oi/Config.cpp
@@ -13,7 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#include "oi/OIUtils.h"
+#include "oi/Config.h"
 
 #include <glog/logging.h>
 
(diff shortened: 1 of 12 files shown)


oi/Config.h
```diff
@@ -15,18 +15,21 @@
...
 
-namespace oi::detail::utils {
+namespace oi::detail::config {
 
-std::optional<FeatureSet> processConfigFile(const std::string& configFilePath,
-                                            std::map<Feature, bool> featureMap,
-                                            OICompiler::Config& compilerConfig,
-                                            OICodeGen::Config& generatorConfig);
+std::optional<FeatureSet> processConfigFiles(
+    std::span<const std::filesystem::path> configFilePath,
```
ajor, line 30: nit:
```suggestion
    std::span<const std::filesystem::path> configFilePaths,
```

oi/Config.cpp
```diff
@@ -27,16 +27,48 @@
...
+  FeatureSet disables;
+
+  for (fs::path p : configFilePaths) {
+    auto fs = processConfigFile(p, compilerConfig, generatorConfig);
+    if (!fs.has_value())
+      return std::nullopt;
+    enables |= *fs;
+  }
+
+  // Override anything from the config file with command line options
```
ajor, line 57:
```suggestion
  // Override anything from the config files with command line options
```

test/integration/runner_common.cpp
```diff
@@ -136,55 +136,15 @@ int IntegrationBase::exit_code(Proc& proc) {
...
-  // If no extra config provided, return the config path unaltered.
-  if (prefix.empty() && suffix.empty()) {
-    return configFile;
-  }
-
-  auto customConfigFile = workingDir / "oid.config.toml";
-  auto config = toml::parse_file(configFile);
-
-  // As relative paths are allowed, we must canonicalise the paths before
-  // moving the file to the temporary directory.
```
ajor, line 150: Is this canonicalisation not still needed?

ajor (approved):

Issue #359, opened 2023-09-21
issues with CMake dependency handling


milianw:
Hey all,

I'm trying to build OI on arch and run into a lot of trouble.

- the rocksdb version you reference doesn't support libstdc++ from gcc 13, giving compile errors due to missing cstdint includes
- you hardcode LLVM 15, but I have LLVM 16
- you have a mixed bag of system libs, and FetchContent, and  git submodules. Why is that? I'm all for system libs, unless you need to have a very specific version of your thirdparty libs - is that the case? If so, would you be opposed to using submodules everywhere? The big advantage there is that we can more easily patch things, and also share the downloaded code across multiple build folders (think debug vs. release)
- your msgpack CMake hackery doesn't work anymore with msgpack-cxx 5.0.0

I have workarounded all of these issues locally now, and would like to clean things up a bit upstream. But for that I would need to know what your constraints are.

My suggestion would be:

- replace FetchContent with git submodules
- then update rocksdb to a newer versions that works with newer compilers.
- alternatively (additionally?) we could use `find_package` for more dependencies. all of the following are covered by system packages on arch:
    - tomlplusplus
    - glog
    - rocksdb
- support newer msgpack-cxx and only if that's not found fallback to the existing cmake hackery.

Would you be OK with patches going in direction?

milianw:
OK, LLVM15 is available on my system after all, so updating that isn't that important for now after all. There are some code changes required for LLVM16, like:

```
@ oi/OICompiler.cpp:242 @ class OIMemoryManager : public RTDyldMemoryManager {
  bool needsToReserveAllocationSpace(void) override {
    return true;
  }
 
  void reserveAllocationSpace(
      uintptr_t, uint32_t, uintptr_t, uint32_t, uintptr_t, uint32_t) override;
      uintptr_t, Align, uintptr_t, Align, uintptr_t, Align) override;

  uint8_t* allocateCodeSection(uintptr_t,
                               unsigned,
```

and

```
@ oi/OICompiler.cpp:294 @ void OIMemoryManager::reserveAllocationSpace(uintptr_t codeSize,
  uint64_t totalSz = alignTo((codeSize + roDataSize + rwDataSize), 1024);

  VLOG(1) << "reserveAllocationSpace: codesize " << codeSize << " codeAlign "
          << codeAlign << " roDataSize " << roDataSize << " roDataAlign "
          << roDataAlign << " rwDataSize " << rwDataSize << " rwDataAlign "
[...]

Issue #114, opened 2023-03-29
Warnings from C++ dependencies during build


JakeHillion:
We currently get warnings from `glog` during our build. This is because `FetchContent` targets are built as opposed to treated as external.

When we update `cmake_minimum_required(VERSION 3.25)` we can add `SYSTEM` to each `FetchContent_Declare` which will correct this behaviour. Adding this issue to track it, as now is currently too early to update for most distributions (https://repology.org/project/cmake/versions).

PR #373, opened 2023-09-29
FetchContent Improvements

milianw:
## Summary

This patch series makes it possible to consume more library from the system, by leveraging the new ` FIND_PACKAGE_ARGS` facility for `FetchContent`.

With the way msgpack is now included, I can now finally compile this repo without any extra patches.

## Test plan

It compiles :)

Fixes: https://github.com/facebookexperimental/object-introspection/issues/359

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -39,6 +39,7 @@ FetchContent_Declare(
   GIT_REPOSITORY https://github.com/marzer/tomlplusplus.git
   GIT_TAG        4b166b69f28e70a416a1a04a98f365d2aeb90de8 # v3.2.0
   GIT_PROGRESS   TRUE
+  FIND_PACKAGE_ARGS 3.2.0 CONFIG
 )
 FetchContent_MakeAvailable(tomlplusplus)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,23 +78,34 @@ FetchContent_Declare(
 FetchContent_MakeAvailable(googletest)
 
 ### rocksdb
-FetchContent_Declare(
-  rocksdb
-  SYSTEM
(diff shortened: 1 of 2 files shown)


JakeHillion (commented):
First comments: looks good to me except for CMake 3.25. Arch and a few other distros have it, but the CentOS we use internally and the Ubuntu used in the CI don't. We have an issue tracking this as the `SYSTEM` parameter should definitely be added when it's more widespread (#114).

Appreciate the efforts! This doesn't build on my work machine with the version put back to 3.20 as is (mostly our fault) so I need a bit of time to get to the bottom of that.

milianw:
@JakeHillion while `SYSTEM` requires CMake 3.25, removing that won't bring a substantial advantaged since `FIND_PACKAGE_ARGS` requires CMake 3.24 - see https://cmake.org/cmake/help/latest/module/FetchContent.html From what I gather, that version is still too new for you?

If so, we need to take a different approach, one more akin to what I did for RocksDB in https://github.com/facebookexperimental/object-introspection/pull/373/commits/5a6ef38c376a7ec56c748f76fef647258408c92e - Are you OK with that?

JakeHillion:
Sorry for not picking up on that. The approach taken with RocksDB seems good (if !find_package then fetchcontent). Could you also create an issue for changing this to `FIND_PACKAGE_ARGS` when 3.24 is widely available, and link to it from a comment in the CMake in each case? Then hopefully we'll remember to switch. Thanks again for cleaning this up and making it easier to build OI externally.

JakeHillion (changes requested):
Thanks for looking into this. `if not find_package(...) then fetch_content(...)` seems a good approach to me and should keep us compatible.

PR #376, opened 2023-10-08
ci: move formatting checks to nix

JakeHillion:
ci: move formatting checks to nix


ajor (approved):
Can't say I understand it, but as long as it works, let's go!

PR #383, opened 2023-10-13
tbv2: improve equality for iterator

JakeHillion:
## Summary

The iterator in treebuilder v2 currently has an equality function which only correctly compares `end()` iterators. Extend the iterator with a count of the number of times it has been incremented, which combined with `data_` and whether it is complete (lack of `next_`) provides equality on elements which aren't end.

This isn't really a functional change - comparisons to `end()` still work as they did before, which is the only real usage in an iterator.

## Test plan

- CI


include/oi/IntrospectionResult-inl.h
```diff
@@ -23,6 +23,7 @@ static_assert(false,
 #define INCLUDED_OI_INTROSPECTIONRESULT_INL_H 1
 
 #include <cassert>
+#include <limits>
```
ajor, line 26: Not needed?

ajor (approved):

PR #385, opened 2023-10-13
capture_keys: store dynamic type path components more efficiently

JakeHillion:
capture_keys: store dynamic type path components more efficiently


ajor (approved):
Looks good!

PR #392, opened 2023-10-25
TopoSorter: Only allow certain params to be incomplete

ajor:
For the containers which are allowed to be declared with incomplete types, it is only the contained types which are allowed to be incomplete. Other template parameters (e.g. allocators) must always be defined before use.

### DIFF
--- a/oi/type_graph/TopoSorter.cpp
+++ b/oi/type_graph/TopoSorter.cpp
@@ -85,7 +85,7 @@ namespace {
  * Other containers are not required to do this, but might still have this
  * behaviour.
  */
-bool containerAllowsIncompleteParams(const Container& c) {
+bool containerAllowsIncompleteParam(const Container& c, size_t i) {
   switch (c.containerInfo_.ctype) {
     case SEQ_TYPE:
     case LIST_TYPE:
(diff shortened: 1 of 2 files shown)


oi/type_graph/TopoSorter.cpp
```diff
@@ -85,30 +85,34 @@ namespace {
...
     case SEQ_TYPE:
     case LIST_TYPE:
     case UNIQ_PTR_TYPE:
     case SHRD_PTR_TYPE:
       // Also std::forward_list, if we ever support that
       // Would be good to have this as an option in the TOML files
-      return true;
+      if (i == 0)
+        return true;
+      return false;
```
JakeHillion, line 98: nit: `return i == 0;`

oi/type_graph/TopoSorter.cpp
```diff
@@ -85,30 +85,34 @@ namespace {
...
       accept(param.type());
     }
   }
   sortedTypes_.push_back(c);
-  if (containerAllowsIncompleteParams(c)) {
-    for (const auto& param : c.templateParams) {
+  for (size_t i = 0; i < c.templateParams.size(); i++) {
+    if (containerAllowsIncompleteParam(c, i)) {
+      const auto& param = c.templateParams[i];
       acceptAfter(param.type());
```
JakeHillion, line 116: doesn't `acceptAfter` mean this only needs one loop? if both cases were merged into the first loop, `accept` would occur immediately but `acceptAfter` would be appropriately delayed. this way might be clearer to read though.

JakeHillion (approved):

PR #395, opened 2023-10-27
codegenv2: improve multi argument generation

JakeHillion:
## Summary

Adds a new function `CodeGen::codegenFromDrgns` which runs multiple drgn root types through one `DrgnParser`. Stores the naming for the output within this function as we previously decided that when doing the actual codegen, which is too late with multiple root types.

This new function is used in `OIGenerator` when you have multiple OIL calls within one input object. Rather than producing separate `.o`s with likely duplicate code for each callsite, we produce a single `.o` that contains both calls. For calls with types with significant overlap this should be a significant reduction in work. The downside is the calls can't have different features, but this could be solved in the future by namespacing the code based on features/config within the generated `.cpp` - the pros seem to outweigh the con.

This should also be used with `oid` in multi probe mode as it would again significantly reduce the work it has to do. I didn't do this yet as it requires changing the cache. As I've got a big cache refactor ongoing at the minute it makes sense to wait until after that's landed to make this change.

## Test plan

- CI for existing behaviour
- [D50835153](https://www.internalfb.com/diff/D50835153) for new behaviour

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -42,6 +42,9 @@
 #include "type_graph/TypeIdentifier.h"
 #include "type_graph/Types.h"
 
+template <typename T>
+inline constexpr bool always_false_v = false;
+
 namespace oi::detail {
 
 using type_graph::AddChildren;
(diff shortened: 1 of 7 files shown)


tyroguru (approved):

JakeHillion:
Going to drop this for now. The rebase was bad enough that I'll rewrite it at some point.

PR #396, opened 2023-10-30
incomplete: handle drgn returning a nullptr name

JakeHillion:
## Summary

The `drgn` calls sometimes set this cstr to `nullptr`. If this happens the implicit `std::string` constructor in `makeType` throws `std::logic_error`. Explicitly check for `nullptr` after the calls and set `"<incomplete>"` accordingly.

## Test plan

\<redacted\> (I tested it I promise)


ttreyer (approved):
I encountered this issue while reviewing this PR... Approved!

PR #397, opened 2023-11-02
TypeGraph: Stop identifying containers in DrgnParser

ajor:
Leave it to the new mutator pass IdentifyContainers to replace Class
nodes with Container nodes where appropriate.

This will allow us to run passes over the type graph before identifying
containers, and therefore before we have lost information about the
internal details of the container (e.g. alignment of member variables).

### DIFF
--- a/oi/type_graph/TypeIdentifier.cpp
+++ b/oi/type_graph/TypeIdentifier.cpp
@@ -86,6 +86,7 @@ void TypeIdentifier::visit(Container& c) {
           }
           c.templateParams[i] = *dummy;
           replaced = true;
+          break;
         }
       }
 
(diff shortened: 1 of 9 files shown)


oi/type_graph/TypeIdentifier.cpp
```diff
@@ -86,6 +86,7 @@ void TypeIdentifier::visit(Container& c) {
           }
           c.templateParams[i] = *dummy;
           replaced = true;
+          break;
```
ajor, line 89: Just a tiny unrelated optimisation I spotted while going through this code

JakeHillion (approved):

PR #398, opened 2023-11-03
collapse TreeBuilderV2 features

JakeHillion:
collapse TreeBuilderV2 features

Summary:

Currently there are two features between CodeGen v2 (TypeGraph) and TreeBuilder
v2. These are TypedDataSegment and TreeBuilderTypeChecking. Each of these
features currently has a full set of tests run in the CI and each have specific
exclusions.

Collapse these features into TreeBuilder v2. This allows for significantly
simplified testing as any OIL tests run under TreeBuilder v2 and any OID tests
run under TreeBuilder v1.

The reasoning behind this is I no longer intend to partially roll out this
feature. Full TreeBuilder v2 applies different conditions to containers than
the intermediate states, and writing these only to have them never deployed is
a waste of time.

Test Plan:
- it builds
- CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -21,20 +21,6 @@ workflows:
           oid_test_args: "-ftype-graph"
           tests_regex: "OidIntegration\\..*"
           exclude_regex: ".*inheritance_polymorphic.*|.*arrays_member_int0"
-      - test:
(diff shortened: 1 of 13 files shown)


ajor (approved):
Looks good!

PR #402, opened 2023-11-13
clang-format: disable bin packing

JakeHillion:
clang-format: disable bin packing

Bin packing often makes code hard to read. Disable it entirely.

Test plan:
- CI


ajor (approved):
Some new annoyances as expected, but overall I'd say this looks like an improvement 👍

PR #410, opened 2023-11-15
tbv2: add dynamic context passed through all functions

### LINKED ISSUE #404: OIL uses a single shared pointer set even though it can be called concurrently
The `static pointers` set is used by both OIL and OID. As OIL can be used concurrently this is an issue. Fix by having one pointer set per request which is fine because OIL can allocate. This will require passing more context around through every function.

JakeHillion:
tbv2: add dynamic context passed through all functions

Previously for we had some shared state between all requests, noticeably the
pointers set. This change adds a by reference value to all requests which can
hold additional mutable state. The pointers set is moved into this mutable
state for OIL, which means each concurrent request will have its own pointer
set. Doing things this way allows more features to be added in the future
without such a big code modification.

Closes #404

Test plan:
- CI

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -128,8 +128,9 @@ void addIncludes(const TypeGraph& typeGraph,
     includes.emplace("oi/types/st.h");
   }
   if (features[Feature::Library]) {
-    includes.emplace("vector");
+    includes.emplace("memory");
     includes.emplace("oi/IntrospectionResult.h");
+    includes.emplace("vector");
   }
   if (features[Feature::JitTiming]) {
     includes.emplace("chrono");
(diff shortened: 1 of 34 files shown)


ajor (approved):
It'd be nice if we could have a constant representing the size of the PointerHashSet
(`1 << 20) / sizeof(uintptr_t)`) instead of duplicating it in a range of places. I'd understand if it's not that easy because of the split between CodeGen and FuncGen though.

JakeHillion:
> It'd be nice if we could have a constant representing the size of the PointerHashSet (`1 << 20) / sizeof(uintptr_t)`) instead of duplicating it in a range of places. I'd understand if it's not that easy because of the split between CodeGen and FuncGen though.

Yep, it's a bit of a pain. I was planning to make it variable for OIL at some point but there's still some duplication. I think I'll add it as a default to the template parameter, then the default won't need to be specified but we'll still be able to vary it in future.

PR #411, opened 2023-11-15
Make pointer tracking array facility fail gracefully when it is full

tyroguru:
## Summary

The pointer tracking mechanism we have doesn’t handle full occupancy. It keeps looking for new entries for ever so the JIT thread ends up spinning which obviously extremely bad.

The current fix is to keep a count of the number of entries added and that equals the size of the array we return false. This normally indicates to the consumer that we have seen this pointer before but obviously this is not the case. What it means is that the results are partial and not complete. To indicate that this is the case we just store the number of entries in the array and its capacity in the dataHeader and display this at the end. If they are equal then warn that the results are partial. The current changes do not include making the array size unable - this really needs doing. For the minute I have left the implementation exhausting the array which is extremely CPU inefficient but makes maximal use of space. The whole thing probably needs revisiting and optimising. It definitely needs to be tunable.

As we can't currently size the array I manually modified the size of the  ‘pointers’ array to by 128 pointers (1<<10 / sizeof(uintptr_t). I

## Test plan

As we can't currently size the array I manually modified the size of the  ‘pointers’ array to by 128 pointers (1<<10 / sizeof(uintptr_t). I then added 1024 `std::unique_ptr<std::string` pointers into a vector and introspected it. We see the following in the logs at -d 1:

```
V1115 04:12:24.883224 2246927 OIDebugger.cpp:2904] Pointer tracking stats: 128/128
V1115 04:12:24.883231 2246927 OIDebugger.cpp:2908] Pointer tracking array is exhausted! Results may be partial.
```

A 'make test' has no new failures.

Check that OI is working correctly, you can add some unit or integration test,
paste the output of some manual test, and / or paste the output of running
the test locally with `make test-static`

### DIFF
--- a/oi/OITraceCode.cpp
+++ b/oi/OITraceCode.cpp
@@ -70,8 +70,8 @@ class {
   /*
    * Adds the pointer to the set.
    * Returns `true` if the value was newly added. `false` may be returned if
-   * the value was already present or if there are no entries left in the
-   * array.
+   * the value was already present, a null pointer was passed or if there are
+   * no entries left in the array.
    */
   bool add(uintptr_t pointer) noexcept {
     if (pointer == 0) {
--- a/oi/OITraceCode.cpp
+++ b/oi/OITraceCode.cpp
@@ -44,9 +44,12 @@ class {
  private:
   // 1 MiB of pointers
   std::array<uintptr_t, (1 << 20) / sizeof(uintptr_t)> data;
+  size_t numEntries;
 
-  // twang_mix64 hash function, taken from Folly where it is used
-  // as the default hash function for 64-bit integers
+  /*
+   * twang_mix64 hash function, taken from Folly where it is used as the
+   * default hash function for 64-bit integers.
+   */
   constexpr static uint64_t twang_mix64(uint64_t key) noexcept {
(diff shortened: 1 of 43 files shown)


oi/OITraceCode.cpp
```diff
@@ -61,26 +64,43 @@ class {
...
-  // Returns `true` if the value was newly added,
-  // or `false` if the value was already present.
+  /*
+   * Adds the pointer to the set.
+   * Returns `true` if the value was newly added. `false` may be returned if
+   * the value was already present or if there are no entries left in the
+   * array.
+   */
   bool add(uintptr_t pointer) noexcept {
     __builtin_assume(pointer > 0);
```
JakeHillion, line 77: Could we change this to:
```
if (pointer == 0)
  return false;
```
As a safety feature? It looks to me at the minute that passing `nullptr` would always return true unless the set is full in which case it would return false. It seems like saying always don't track `nullptr` is safer, even though we don't use it like that.

JakeHillion (approved):

oi/OITraceCode.cpp
tyroguru: Yeah, I forgot about that discussion yesterday! Makes much more sense. I'll change it to that.

PR #412, opened 2023-11-15
OID: Don't run tree builder if we are dumping the data segment

ajor:
We generally only dump the data segment when there is some issue running tree builder and we want to process it off-host.

## Test plan
Manual testing:
- Confirmed that tree builder is still run by default and that no data segment is dumped
- Confirmed that tree builder is not run and a data segment is dumped when using "-B"

### DIFF
--- a/oi/OIDebugger.cpp
+++ b/oi/OIDebugger.cpp
@@ -2869,6 +2869,9 @@ bool OIDebugger::processTargetData() {
       if (!dumpDataSegment(req, outVec)) {
         LOG(ERROR) << "Failed to dump data-segment for " << req.arg;
       }
+
+      // Skip running Tree Builder
+      continue;
     }
 
     auto typeInfo = typeInfos.find(req);
@@ -2908,6 +2911,11 @@ bool OIDebugger::processTargetData() {
     if (treeBuilderConfig.features[Feature::GenPaddingStats]) {
       paddingHunter.processLocalPaddingInfo();
     }
+  }
+
+  if (treeBuilderConfig.dumpDataSegment) {
+    // Tree Builder was not run
+    return true;
   }
 
   if (typeTree.emptyOutput()) {


tyroguru (approved):

JakeHillion (changes requested):
Could we add this as a field in `Oid::Config` instead of a feature? It could be added as a `mode` for a similar UX to features, though you can't undo turning a mode on yet. The reason for this is all the features so far directly affect code generation (I realise they are used elsewhere too) and therefore it's been a really clean way to affect the build hash. It would be nice to have the things which exclusively control the running of the debugger in a different set.

tyroguru:
> Could we add this as a field in `Oid::Config` instead of a feature? It could be added as a `mode` for a similar UX to features, though you can't undo turning a mode on yet. The reason for this is all the features so far directly affect code generation (I realise they are used elsewhere too) and therefore it's been a really clean way to affect the build hash. It would be nice to have the things which exclusively control the running of the debugger in a different set.

The ideal way to communicate this IMO is via a mechanism that is attached to a probe in the script - this has been talked about before. In that way the state travels with the probe specification. I had completely forgot that was use the features to control cache generation and obviously we don't want to trigger a rebuild for something that doesn't affect the generated BLOB.

JakeHillion:
> > Could we add this as a field in `Oid::Config` instead of a feature? It could be added as a `mode` for a similar UX to features, though you can't undo turning a mode on yet. The reason for this is all the features so far directly affect code generation (I realise they are used elsewhere too) and therefore it's been a really clean way to affect the build hash. It would be nice to have the things which exclusively control the running of the debugger in a different set.
> 
> The ideal way to communicate this IMO is via a mechanism that is attached to a probe in the script - this has been talked about before. In that way the state travels with the probe specification. I had completely forgot that was use the features to control cache generation and obviously we don't want to trigger a rebuild for something that doesn't affect the generated BLOB.

A command line flag that only affects the debugger and not the cache blob seems good. A change to the script would also be fine, but we'd have to be careful about caching too then - currently we'd be able to hash the entire script text/parsed script and get a decently useful thing for the hash. With pragmas that affect how the debugger runs and not codegen we'd have to be a lot more careful. Something to think about longer term.

ajor:
Updated to avoid using feature flags

PR #420, opened 2023-11-23
TypeGraph: Fix handling for classes which inherit from containers

ajor:
TypeGraph: Fix handling for classes which inherit from containers

We previously moved container identification later in CodeGen in order
to preserve information for AlignmentCalc.

However, Flattener needs to know if a class is a container in order to
apply its special handling for this case.

This new approach moves container identification in front of Flattener,
but has Container own a type node, representing its layout. This
underlying type node can be used for calculating a container's
alignment in a later pass.

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -1157,9 +1157,9 @@ void CodeGen::transform(TypeGraph& typeGraph) {
 
   // Simplify the type graph first so there is less work for later passes
   pm.addPass(RemoveTopLevelPointer::createPass());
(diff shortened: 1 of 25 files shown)


JakeHillion (approved):
Looks good! Could we have `IdentifyContainers` back with this new approach as a very early pass? I think the complexity of the mutator is a fair bit worse than doing it in the parser, but it would reduce duplication as we're writing more parsers at the minute. Your call.

ajor:
Updated to restore IdentifyContainers pass

PR #422, opened 2023-12-07
Handle function drgn type when enumerating types

tyroguru:
## Summary
Handle function types when enumerating types.

## Test plan
A test needs adding.

### DIFF
--- a/oi/type_graph/DrgnParser.cpp
+++ b/oi/type_graph/DrgnParser.cpp
@@ -116,6 +116,9 @@ Type& DrgnParser::enumerateType(struct drgn_type* type) {
       case DRGN_TYPE_VOID:
         t = &enumeratePrimitive(type);
         break;
+      case DRGN_TYPE_FUNCTION:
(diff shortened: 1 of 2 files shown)


JakeHillion:
I'm wondering if this change makes total sense - we're taking in a function type which is explicitly unsized, and stating that it's a sized type `StubbedPointer`. I wonder if it would make more sense to immediately return an `Incomplete` from this switch case, along with a depth decrement. That would work if the goal is to prevent hitting the `catch` statement. I could be missing something though!

I wrote this bit of code which I think neatens up the depth decrement, if we do decrement in this specific switch case it would make sense to use something like this: https://github.com/JakeHillion/object-introspection/blob/6be7e7708bfd69c75faa46d42df81a79734eab46/oi/type_graph/ClangTypeParser.cpp#L45-L55

ajor:
Finding a test case for this proved difficult. Putting this on hold until we hit this issue in production again.

PR #423, opened 2023-12-07
Update Python used by CMakeLists.txt

ttreyer:
## Summary
Add a new Python version used by our CMakeLists.txt.
Also, ensure that `integration_test_target` always has debug information, even for Release build.

## Test plan
```
$ make test
```


JakeHillion (commented):
This doesn't work on my machine... I get `ModuleNotFoundError: No module named 'toml'`

PR #424, opened 2023-12-11
Add skeleton LLDBParser

ttreyer:
## Summary
Add the bare-bones class for LLDBParser, with just the public API throwing "unimplemented" exception.
Duplicate the DRGNParser test suite and adapt it for LLDBParser.

## Test plan
```
$ make test
```


JakeHillion:
This PR has quite a few commits now, would you like to split out the skeleton and land that by itself or wait until it's feature complete? Up to you!

PR #428, opened 2023-12-13
Add support for Simple Template Names

ttreyer:
## Summary
The `.debug_string` table can get very large for massive programs. If it grows larger than 4GiB, DWARF is unable to address any string beyond the limit, as it uses `uint32_t` to store an offset within the table. But the table also has lots of duplication: every templated types' name contains all the template parameters too. Knowing that, the option `-gsimple-template-names` saves space by only storing the template type's name, without its template parameters. The complete name can then be reconstructed by traversing the DIEs for the template parameters, so that no information is lost.

This PR imports Omar's work on drgn to add support for Simple Template Names.

## Test plan
```
$ make test
```


JakeHillion (approved):
let's do it :)

PR #429, opened 2023-12-13
type_graph: Add reference type

JakeHillion:
type_graph: Add reference type

Our existing drgn parser emits pointers, regardless of whether the type in the
DWARF is really a reference. Newer parsers like the LLDB/Clang parsers can
differentiate between a pointer and a reference. We should do this as
references can be safely followed according to the language rules with
`ChaseRawPointers` disabled, while pointers cannot.

Test Plan:
- TBD

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebookexperimental/object-introspection/pull/429).
* #421
* __->__ #429

### DIFF
--- a/oi/type_graph/NameGen.cpp
+++ b/oi/type_graph/NameGen.cpp
@@ -200,6 +200,14 @@ void NameGen::visit(Pointer& p) {
   p.setInputName(inputName);
 }
 
+void NameGen::visit(Reference& p) {
+  RecursiveVisitor::visit(p);
+  p.regenerateName();
+  std::string inputName{p.pointeeType().inputName()};
+  inputName += '*';
+  p.setInputName(inputName);
+}
+
 void NameGen::visit(DummyAllocator& d) {
   RecursiveVisitor::visit(d);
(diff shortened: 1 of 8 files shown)


JakeHillion:
Started testing this and it makes some dodgy assumptions. First the size in tbv2 comes out wrong because `sizeof(int&) == sizeof(int) != sizeof(int*)`. Secondly if we follow references in codegen we don't check for cycles as we assume it's already safe, and because reference cycles are possible (if unlikely) this makes this generation dangerous.

The correct approach is to generate these as pointers in the generated code (regardless of ChaseRawPointers) and name them as references. Perhaps it would make more sense to do this as part of `Pointer` as it's now less different, but I think we should be liberal with Type names to make our intentions clearer - clang for example has many only slightly different types.

I think I'll squash this into the ClangTypeParser changes so it can be tested properly, as unit tests can't cover our codegen cleanly and there's currently no way to get a Reference type out of DrgnParser.

PR #431, opened 2023-12-13
Type Graph: Replace MutationTracker with the more general ResultTracker

ajor:
MutationTracker could only store Type nodes, while ResultTracker is templated on the result type so can store anything.
    
Template the Visitor base class on the return type of visit() functions.
    
This sets us up for allowing visitors to return different results from their visit() functions in the future.

This will be used in a future commit introducing DrgnExporter, where we cache drgn_type* results while walking the type graph: #425

### DIFF
--- a/oi/type_graph/NodeTracker.h
+++ b/oi/type_graph/NodeTracker.h
@@ -109,23 +109,27 @@ class NodeTrackerHolder {
 };
 
 /*
- * MutationTracker
+ * ResultTracker
  *
- * Helper class for mutators. Efficiently tracks visited and replaces nodes.
+ * Efficiently caches the results of visited nodes.
  */
-class MutationTracker {
+template <typename T>
+class ResultTracker {
  public:
-  MutationTracker(size_t size) : visited_(size) {
+  ResultTracker() = default;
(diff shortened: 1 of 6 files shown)


oi/type_graph/NodeTracker.h
```diff
@@ -109,23 +109,25 @@ class NodeTrackerHolder {
...
+  ResultTracker() = default;
+  ResultTracker(size_t size) : visited_(size) {
   }
 
   /*
    * get
    *
-   * Returns a type pointer if the given node has been visited or replaced.
+   * Returns the cached result if the given node has been visited.
    * Returns nullptr if this node has not yet been seen.
```
JakeHillion, line 127: s/nullptr/a default constructed element/?

JakeHillion (approved):

oi/type_graph/NodeTracker.h
ajor: I hadn't actually changed enough code to handle arbitrary types - I've added a static_assert that the result type is a pointer now. I'll leave handling arbitrary types as a future extension.

PR #432, opened 2023-12-14
NameGen: Override inputName for anonymous members

### LINKED ISSUE #292: Stubbed types do not appear in the generated code as padding
In v1 OID would report fields like `__anon_member_0` when there is an anonymous union that is a member of a struct. In type graph (as I understand it) these unions are completely removed from the graph. This means when the information necessary to treebuild is generated in codegen they appear as padding and thus aren't available in the output.

I think the solution to this is generating the fields of stubbed types as `Dummy` with their own name rather than a padding array. This should then generate the correct output.

Examples:
- `OilIntegration.anonymous_anon_union`

ajor:
Fixes #292

### DIFF
--- a/test/integration/anonymous.toml
+++ b/test/integration/anonymous.toml
@@ -192,7 +192,6 @@ definitions = '''
     }]'''
 
   [cases.anon_union]
-    oil_skip = "anonymous unions are fully stubbed in the generated code" # https://github.com/facebookexperimental/object-introspection/issues/292
     param_types = ["const AnonUnionContainer&"]
     setup = 'return AnonUnionContainer{ .a = 3 };'
     cli_options = ["-fchase-raw-pointers"]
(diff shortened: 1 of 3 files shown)


test/integration/anonymous.toml
```diff
@@ -209,8 +208,8 @@ definitions = '''
       "staticSize": 24,
       "exclusiveSize": 10,
       "members": [
-        {"name":"__anon_member_0", "staticSize":2, "exclusiveSize":2},
-        {"name":"__anon_member_1", "staticSize":8, "exclusiveSize":8},
+        {"name":"__oi_anon_0", "staticSize":2, "exclusiveSize":2},
+        {"name":"__oi_anon_2", "staticSize":8, "exclusiveSize":8},
```
ajor, line 212: A bit annoying that this is "_2" not "_1", but it's because of padding members. Didn't think it was worth fixing right now

JakeHillion (approved):

PR #433, opened 2023-12-14
OID: Make CodeGen v2 (TypeGraph) the default

ajor:
##
Set CodeGen v2 as the default mode. CodeGen v1 can still be requested with "-Ftype-graph".

Update the CI to only run OID tests with v2. OIL tests still run with v1. Rename the CI jobs to reflect that v2 is now the default and v1 is the special-case.

## Test plan
- CI
- Run oid manually without specifying "-ftype-graph" and confirm that CodeGen v2 is run


JakeHillion (approved):
hype. you need to change the required checks in the repo settings with the renaming.

PR #443, opened 2023-12-22
type_graph: avoid overwriting explicitly set alignment

JakeHillion:
type_graph: avoid overwriting explicitly set alignment

Previously AlignmentCalc calculates the alignment and sets packing for every
type except a member with explicit alignment. Change this to check whether an
alignment has been previously set for a type before calculating it. Use this in
ClangTypeParser where the full alignment of the type is available.

Remove explicitly aligning members by the type because that was previously
reserved for members with explicit alignment. AlignmentCalc will correctly
align a member to the underlying type without this. Explicit member alignment
is still missing, as before this change.

Test plan:
- CI
- Too little. Gets further into a production type than without this change.

### DIFF
--- a/oi/type_graph/AlignmentCalc.cpp
+++ b/oi/type_graph/AlignmentCalc.cpp
@@ -50,23 +50,25 @@ void AlignmentCalc::accept(Type& type) {
 void AlignmentCalc::visit(Class& c) {
   RecursiveVisitor::visit(c);
 
-  uint64_t alignment = 1;
-  for (auto& member : c.members) {
-    if (member.align == 0) {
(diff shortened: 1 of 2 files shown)


ajor (approved):
No tests for ClangTypeParser? 😢

JakeHillion:
> No tests for ClangTypeParser? 😢

Unfortunately not... I was hoping to enable all the integration tests and get coverage that way then add unit tests later. The integration tests PR is stalled because of the build system stuff so I might look into adding unit tests similar to DrgnParser next week - it's also not trivial though because it needs the compilation command of the integration test target and not just the compiled binary from CMake.

PR #447, opened 2023-12-29
circleci: clean up codegen v1 runs

JakeHillion:
circleci: clean up codegen v1 runs

Remove the CodeGen v1 sections of the CI config because both OID and OIL use
CodeGen v2.

We were missing running any test that wasn't `Oi{d,l}Integration.*` before.
This now runs the unit tests again and requires a minor fix to one unit test.

Test plan:
- CI


JakeHillion:
I'll remove the requirements on `test-codegenv1-clang` and `test-codegenv1-gcc` when this is approved.

tyroguru (approved):

ajor (approved):

Issue #295, opened 2023-08-18
CodeGen v2 codegen fails for zero length array


JakeHillion:
JIT compiler error:
```
1925988647140263860.cc:138:7: error: base class has incomplete type
    : validate_size<Type, ExpectedSize, sizeof(Type)> {};
      ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
1925988647140263860.cc:213:15: note: in instantiation of template class 'validate_size<OIInternal::(anonymous namespace)::Foo0_0, 0, 0>' requested here
static_assert(validate_size<Foo0_0, 0>::value);
              ^
1925988647140263860.cc:132:8: note: definition of 'validate_size<OIInternal::(anonymous namespace)::Foo0_0, 0, 0>' is not complete until the closing '}'
struct validate_size : std::true_type {
       ^
1 error generated.
```

Unclear why this is happening. #294 does not fix it, and everything looks to be defined in the correct order. Needs further debugging.

Examples:
- `build/test/integration/integration_test_runner --gtest_filter='OidIntegration.arrays_member_int0' --verbose --preserve --enable-feature type-graph`

ttreyer:
Why do we have both a 3-template-params version and a 2-template-param version that inherit from the first? Do we ever use the 3-template-params version directly?

I'm able to pass the static assert if I use two distinct structs (like `validate_size_3` and `validate_size_2`), instead of only one version that inherit from the other. Both Clang and GCC are unhappy about the inheritance, but fine with the two distinct structs.
See https://godbolt.org/z/9YWMvozof, I could update our implementation to use the two distinct structs, but I don't know if CodeGen will need updating.

JakeHillion:
I see no problem with renaming the 3 param version, as you say it should only be used indirectly by the 2 param version. The reason there are two is to get the ActualSize explicitly in the 3 parameter version so it's included in the compiler error.

The best fix is probably to rename the two param version to `validate_size_eq` and remove the default value on `ActualSize`.

Thanks for looking at this!

ttreyer:
Fixed by #453

PR #453, opened 2024-01-10
Fix CodeGen V2 for zero-length array 

ttreyer:
## Summary
Fix issue #295
1. Replace the recursive `validate_size` template with a non-recursive one, making CodeGen compile
2. Process zero-length array as containers, so no object size is left unread in the data buffer
3. Enable the zero-length array test

Is it ok to process zero-length array as container?
I had to modify the test's results to all `0` sizes. Is this the behavior expected? Or should it report an exclusive size of `1` for the `Foo` struct?

## Test plan
The example provided in #295 now passes successfully and has been enabled in the integration tests.
```
$ build/test/integration/integration_test_runner --gtest_filter='OidIntegration.arrays_member_int0' --verbose --preserve --enable-feature type-graph
```

### DIFF
--- a/test/integration/arrays.toml
+++ b/test/integration/arrays.toml
@@ -42,10 +42,6 @@ definitions = '''
         "capacity":10
       }]}]'''
   [cases.member_int0]
-    oil_skip = 'zero length arrays fail codegen v2' # https://github.com/facebookexperimental/object-introspection/issues/295
-    # WARNING: zero-length arrays are handled differently to non-empty arrays.
-    # They end up not being treated as containers. This should probably change
-    # in the future.
     param_types = ["const Foo0&"]
     setup = "return {};"
     expect_json = '''[{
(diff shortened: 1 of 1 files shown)


oi/TreeBuilder.cpp
```diff
@@ -379,8 +379,7 @@ uint64_t TreeBuilder::next() {
 
 bool TreeBuilder::isContainer(const Variable& variable) {
   return th->containerTypeMap.contains(variable.type) ||
-         (drgn_type_kind(variable.type) == DRGN_TYPE_ARRAY &&
-          drgn_type_length(variable.type) > 0);
+         drgn_type_kind(variable.type) == DRGN_TYPE_ARRAY;
```
JakeHillion, line 382: This change breaks CodeGen-v1 for zero length arrays: `make oid && OID_TEST_ARGS='-Ftype-graph' build/test/integration/integration_test_runner --verbose --preserve --force --gtest_filter="OidIntegration.arrays_member_int0"`

I think this works instead:
```cpp
bool TreeBuilder::isContainer(const Variable& variable) {
  if (th->containerTypeMap.contains(variable.type))
    return true;
  if (drgn_type_kind(variable.type) != DRGN_TYPE_ARRAY)
    return false;
  return config.features[Feature::TypeGraph] || drgn_type_length(variable.type) > 0;
}
```

JakeHillion (commented):
Not sure how much we care about breaking codegen-v1 at this point, but this does seem to. Pretty easy to make it work with both :) if we have consensus that codegen-v1 is dead that I missed then you don't need to worry.

JakeHillion (approved):

PR #454, opened 2024-01-10
Add namespaces to integration tests' "typeName"

ttreyer:
## Summary
Some integration tests have not been updated since drgn started reporting type names with all their namespaces, and were failing.

## Test plan
Running the tests now report 15 failure, instead of 32 previously. The polymorphic tests are still failing due to TypeGraph not supporting them yet.


JakeHillion (approved):
Nice! Thanks for catching this. I don't think there's much point trying to enable the polymorphic ones in the CI because as you say lots of them don't work at the minute. We could try enabling the `arrays_member_int0` one you fixed yesterday though.

PR #455, opened 2024-01-11
tbv2: correctly account for list overhead

JakeHillion:
tbv2: correctly account for list overhead

`std::list` has per element overhead for the individual heap allocations. This
was already calculated in the container implementation but not used. Allocate
the overhead of each element in the `std::list` to the `std::list` itself as
with other sequential containers.

Test Plan:
- CI
- Updated test cases

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/facebookexperimental/object-introspection/pull/455).
* #452
* #456
* __->__ #455

### DIFF
--- a/test/integration/std_list.toml
+++ b/test/integration/std_list.toml
@@ -18,7 +18,7 @@ definitions = '''
     param_types = ["const std::list<int>&"]
     setup = "return {{1,2,3}};"
     expect_json = '[{"staticSize":24, "dynamicSize":12, "length":3, "capacity":3, "elementStaticSize":4}]'
-    expect_json_v2 = '''[{"staticSize":24, "exclusiveSize":24, "size":36, "length":3, "capacity":3, "members":[
+    expect_json_v2 = '''[{"staticSize":24, "exclusiveSize":84, "size":96, "length":3, "capacity":3, "members":[
       {"staticSize":4, "exclusiveSize":4, "size":4},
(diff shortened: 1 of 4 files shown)


ttreyer (commented):
Is that consistent with our implementations for other types?
For instance, for `std::set` in `types/set_type.toml`, we have no indirection and directly go on to process the child. The exclusive size then aggregates the overhead of each element, see  [`types/set_type.toml:114`](https://github.com/facebookexperimental/object-introspection/blob/main/types/set_type.toml#L114), and ensure correct reporting of the overhead.

JakeHillion:
> Is that consistent with our implementations for other types? For instance, for `std::set` in `types/set_type.toml`, we have no indirection and directly go on to process the child. The exclusive size then aggregates the overhead of each element, see [`types/set_type.toml:114`](https://github.com/facebookexperimental/object-introspection/blob/main/types/set_type.toml#L114), and ensure correct reporting of the overhead.

Thanks for pointing that out! Thinking about it I agree. While specifically `std::list` suits this approach because each element has very specific overhead, no other lists do. `std::deque` can't split the overhead neatly between elements, and `std::vector` has no concept of per element overhead. I'll change this to attribute the overhead of the elements to the container as you say.

ttreyer (approved):
LGTM 🙂

PR #457, opened 2024-01-11
tbv2: remove unnecessary copy in Element

JakeHillion:
tbv2: remove unnecessary copy in Element

`IntrospectionResult::const_iterator` iterates through the `Element`s in an
`IntrospectionResult`. `Element` currently copies the `type_path` which is a
`std::vector<string_view>` every time the iterator is incremented. This is
unnecessary as the data in the vector only changes slightly between iterations.

This change changes the `type_path` field in `Element` to a
`std::span<const std::string_view>`. Doing this previously caused SEGVs because
of the iterator's potential to be copied. To make it possible we do two things:
1. Make all copies explicit using a clone interface as in `ContainerInfo`. This
   prevents accidental copies of an expensive structure.
2. After calling the copy constructor in `clone()` update the `span` in `next_`
   to point at the newly copied structure.

Moves are fine because the `span` points at the allocation of the `vector`, not
the vector itself.

Test plan:
- CI
- `FILTER='OilIntegration.*' make test`
- Ran `OilgenIntegration.std_vector_vector_int_some` which SEGVd with the
  `span` change before and now doesn't. This now passes cleanly with ASAN
  enabled on the target, though isn't available in `main` (only works on my
  machine).

### DIFF
--- a/include/oi/IntrospectionResult-inl.h
+++ b/include/oi/IntrospectionResult-inl.h
@@ -45,7 +45,9 @@ inline IntrospectionResult::const_iterator IntrospectionResult::begin() const {
   return cbegin();
 }
 inline IntrospectionResult::const_iterator IntrospectionResult::cbegin() const {
-  return ++const_iterator{buf_.cbegin(), inst_};
+  auto it = const_iterator{buf_.cbegin(), inst_};
+  ++it;
+  return it;
 }
 inline IntrospectionResult::const_iterator IntrospectionResult::end() const {
   return cend();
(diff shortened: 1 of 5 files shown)


include/oi/IntrospectionResult-inl.h
```diff
@@ -70,6 +72,18 @@ inline const result::Element* IntrospectionResult::const_iterator::operator->()
...
 }
 
+inline IntrospectionResult::const_iterator
+IntrospectionResult::const_iterator::clone() const {
+  auto ret{*this};
+
+  // Fix the self referential type_path_ field in next_
+  if (ret.next_.has_value()) {
+    ret.next_->type_path = ret.type_path_;
+  }
```
JakeHillion, line 82: I wouldn't mind if this was in the private copy constructor but I don't want to lose the `= default` for it. I don't think there's any way to say use the default then run an extra bit of code directly in the copy constructor - please correct me if I'm wrong.

tyroguru (approved):

PR #461, opened 2024-01-15
tbv2: fix pointer codegen

### LINKED ISSUE #458: tbv2: pointer code generation is broken
Code generation for raw pointers/references is broken in TreeBuilder-v2. This was missed before because the chase pointer tests are all disabled for OIL, while only the tests that chase bad pointers need to be disabled.

Examples:

JakeHillion:
tbv2: fix pointer codegen

A previous change enabled running OIL tests with specific features enabled.
This highlighted that pointer code generation under TreeBuilder-v2 was very
broken. This change updates pointer code generation to work and enables the
skipped tests. All enabled tests need `expected_json_v2` added to them due to
formatting differences.

Reformatted and rewrote the basic type handler that handles primitives and
pointers. Removed the reliance on `features` to decide whether to generate for
TreeBuilder-v2 as the intermediate features have been removed. Pointers are
treated as containers with a capacity of 1 and a length of 0 if null/a cycle
and 1 if followed. This holds for void pointers where, although they aren't
followed, the length is still set.

There were a couple of other changes needed to enable these tests on TBv2 that
aren't worth their own issues and PRs, I sneaked them in here.

Extra changes:
- Added `Pointer` and `Reference` to TopoSorter so they generate
  `NameProvider` instances. It might be worth visiting the graph differently
  for `NameProvider` as it requires so many instances that others generators do
  not. Will consider that in the future.
- Follow typedefs when calculating exclusive size for a type.

Closes #458.

Test plan:
- CI
- Enabled previously disabled tests.

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -195,7 +195,12 @@ void genDecls(const TypeGraph& typeGraph, std::string& code) {
 namespace {
 
 size_t calculateExclusiveSize(const Type& t) {
-  if (const auto* c = dynamic_cast<const Class*>(&t)) {
+  const Type* finalType = &t;
+  while (const auto* td = dynamic_cast<const Typedef*>(finalType)) {
+    finalType = &td->underlyingType();
+  }
+
+  if (const auto* c = dynamic_cast<const Class*>(finalType)) {
     return std::accumulate(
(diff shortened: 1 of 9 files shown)


ajor (approved):
Woohoo!

PR #464, opened 2024-01-19
tbv2: fix Thrift isset lookups with padding

JakeHillion:
tbv2: fix Thrift isset lookups with padding

Thrift isset was failing with a SEGV if the struct contained padding. This is
because we indexed the `isset_indexes` data structure using our field index
rather than the index of the field in Thrift. This then gave a rubbish index
for any exceeding which happens if we have added padding in the middle of the
struct, and this index was looked up in the bitset which can cause a SEGV.

Track a new index `thriftFieldIdx` which is only incremented if we've looked up
a Thrift index.

Namespaced the generated Thrift structs while I was there. This isn't necessary
anymore but cleans things up.

Test plan:
- Added a test case with lots of padding. These don't run in the CI but it
  passes locally.
- `FILTER='OilIntegration.*' make test` - no failures
- `FILTER='OidIntegration.*' make test` - no new failures

### DIFF
--- a/oi/CodeGen.cpp
+++ b/oi/CodeGen.cpp
@@ -505,6 +505,7 @@ void CodeGen::getClassSizeFuncConcrete(std::string_view funcName,
             c.fqName() + ">;\n";
   }
 
+  size_t thriftFieldIdx = 0;
   for (size_t i = 0; i < c.members.size(); i++) {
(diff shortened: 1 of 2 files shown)


arsarwade (approved):
Not really sure how the indexing works, but looks reasonable to me :)

PR #469, opened 2024-01-30
fix string type sso computation

qbojj:
## Summary
The sso flag in string types was calculated wrongly:
```
containter.data() >= &container && containter.data() < (&container + sizeof(container))
```
But because the `&container` is of type `std::basic_string<>*` the + already multiplies by sizeof.

## Note:
Some other types had similar problem, so I have changed them too.

### DIFF
--- a/oi/OITraceCode.cpp
+++ b/oi/OITraceCode.cpp
@@ -185,8 +185,10 @@ struct validate_offset {
 enum class StubbedPointer : uintptr_t {};
 
 bool isStorageInline(const auto& c) {
-  return (uintptr_t)std::data(c) < (uintptr_t)(&c + sizeof(c)) &&
-         (uintptr_t)std::data(c) >= (uintptr_t)&c;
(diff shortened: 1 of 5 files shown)


JakeHillion:
This is great, thanks for taking the time to fix this! Would you consider instead fixing the [`isStorageInline`](https://github.com/facebookexperimental/object-introspection/blob/b5b94ed236e01480bf6706ec2ad8f6bfba350eee/oi/OITraceCode.cpp#L187-L190) helper in `OITraceCode.cpp` and updating these containers to use that check? This would help others avoid this mistake in the future.

qbojj:
I was thinking about making a helper, but (as I don't know the structure of this project) I wasn't sure where to put it. Well as it turns out there was already a helper but no one was using it.

JakeHillion (approved):
Thanks for your contribution! It's a shame we left that bug on the slide but it worked out in the end, and these examples are a big improvement for anyone adding new potentially inline containers.

PR #480, opened 2024-02-07
ignore whole types in addition to members (fully qualified)

tyroguru:
## Summary
Our current mechanism for ignoring members is just that - members only. This change extends this allow stubbing of members by type. 

```
namespace X {
  struct Foo {
    int a, b, c;
  };

  struct Bar {
    int x, y, z;
  };

  struct FooBar {
    Foo f;
    Bar b;
  };
} // namespace X
```

With the above definition to stub out FooBar->Bar you could either stub by member:

```
[[codegen.ignore]]
type="X::Foo"
member = ["b"]
```

or you can stub by type:
```
[[codegen.ignore]]
type="X::Bar"
```

Note this changes the extant behaviour. Now types MUST be fully qualified for them to match. If they are not specified as fully qualified than stubbing won't work as expected and we'll attempt to capture that member/type
.
## Test plan
I've altered the `ignored` integration tests to modify the original test and add 2 more.


oi/type_graph/RemoveMembers.cpp
```diff
@@ -61,18 +61,19 @@ void RemoveMembers::visit(Class& c) {
...
+    if (ignoreMember(c.inputName(), member.name))
       return true;
     if (dynamic_cast<Incomplete*>(&member.type()))
       return true;
     return false;
   });
 }
 
-bool RemoveMembers::ignoreMember(const std::string& typeName,
+bool RemoveMembers::ignoreMember(const std::string_view& typeName,
```
JakeHillion, line 72:
```suggestion
bool RemoveMembers::ignoreMember(std::string_view typeName,
```

oi/type_graph/RemoveMembers.h
```diff
@@ -47,7 +47,7 @@ class RemoveMembers : public RecursiveVisitor {
   void visit(Class& c) override;
 
  private:
-  bool ignoreMember(const std::string& typeName,
+  bool ignoreMember(const std::string_view& typeName,
```
JakeHillion, line 50:
```suggestion
  bool ignoreMember(std::string_view typeName,
```

test/integration/ignored.toml
```diff
@@ -21,18 +26,74 @@ definitions = '''
...
 
+    """
+    expect_json = '''[{
+      "staticSize":96,
+      "exclusiveSize":96,
+      "members":[]
+      }]'''
+
+  [cases.subtype]
+    skip = 'v2 hides the member entirely when it should show it with static size' # todo: github issue
```
JakeHillion, line 39:
```suggestion
```

no need to skip these anymore now v1 is deprecated, if the behaviour changes in the future then the tests should be updated accordingly

test/integration/ignored.toml
```diff
@@ -21,18 +26,74 @@ definitions = '''
...
+        },
+        {
+          "name":"b",
+          "staticSize":96,
+          "exclusiveSize":96
+        }
+      ]}]'''
+
+  [cases.member]
+    skip = 'v2 hides the member entirely when it should show it with static size' # todo: github issue
```
JakeHillion, line 76:
```suggestion
```

JakeHillion (approved):

PR #493, opened 2024-04-16
ClangTypeParser: handle clang::MemberPointer

tyroguru:
## Summary
This diff implements the handling on a clang::MemberPointer. We simply stub the type.

## Test plan

A 'make test` shows no new failures.

### DIFF
--- a/oi/type_graph/ClangTypeParser.cpp
+++ b/oi/type_graph/ClangTypeParser.cpp
@@ -109,6 +109,9 @@ Type& ClangTypeParser::enumerateType(const clang::Type& ty) {
       return enumerateArray(llvm::cast<const clang::ConstantArrayType>(ty));
     case clang::Type::Enum:
       return enumerateEnum(llvm::cast<const clang::EnumType>(ty));
+    case clang::Type::MemberPointer:
+      return enumerateMemberPointer(
+          llvm::cast<const clang::MemberPointerType>(ty));
 
     default:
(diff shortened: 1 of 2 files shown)


oi/type_graph/ClangTypeParser.cpp
```diff
@@ -385,6 +388,11 @@ Type& ClangTypeParser::enumeratePointer(const clang::PointerType& ty) {
   return makeType<Reference>(ty, t);
 }
 
+Type& ClangTypeParser::enumerateMemberPointer(
+    const clang::MemberPointerType& ty) {
+  return makeType<Primitive>(ty, Primitive::Kind::StubbedPointer);
```
ajor, line 393: Thinking about it some more, wouldn't we want to follow a MemberPointer with the same rules as for regular pointers? i.e. never follow function pointers and otherwise check the `chasePointer` value.

I think the case we encountered was for a function pointer(?) so not following it makes sense for now.

Maybe just leave a TODO comment in case we ever come back to this.

ajor (approved):

oi/type_graph/ClangTypeParser.cpp
tyroguru: You're probably right. I'll put a TODO in as suggested but leave it as it is for now. You're right - the case that was failing was a function pointer.

tyroguru: Thanks as always @JakeHillion for the insight!

PR #499, opened 2024-08-12
build: remove dependency on modified folly repository

JakeHillion:
Currently we pull a modified folly repository to be able to use it header only.
The only difference in this repository is adding the `folly-config.h` file.
Pull this patch into this repo instead and use upstream folly.

Bump folly to latest main as well.

Test plan:
- CI build. I have local build issues at the moment.


ajor (approved):
Noice. Any reason not to use a more recent release of folly?

JakeHillion:
> Noice. Any reason not to use a more recent release of folly?

Nope! This is the upstream base of the old modified branch, I'll try the latest hash, should be fine.

PR #502, opened 2024-08-13
nix: add building oid to the flake

JakeHillion:
OI's build is challenging and has often been a problem for the Open Source
community. It requires an extremely specific set of dependencies that are very
hard to achieve on most systems. There are frequent breakages, like when
updating to CentOS Stream 9, or when trying to update the CI's clang from
clang-12 to clang-15 - OI requires the clang libraries to be version 15 but
can't be compiled with it on the CI!

This changes provides a mostly working build environment with `nix`. This
environment is pinned to a specific nixpkgs revision using `flake.lock`, and
only updates when we explicitly tell it to.

Summary of changes:
- Update CMakeLists.txt required version to 3.24. This allows specifying
  `FIND_PACKAGE_ARGS` in `FetchContent`, meaning we can use system packages.
  This is available on most up to date distros (3.30.2 is current).
- Extends `flake.nix` to be able to build OI. Adds instructions for building
  and developing OI using `nix`.
- Partially runs the tests in GitHub Actions. A huge amount must be excluded
  because of incompatibilites between the clangStdenv compiler and drgn. We
  have similar, though fewer, issues when building with the clang-12/libstdcxx
  mix on the Ubuntu 22.04 CircleCI, though this is at least reproducible.
- Updates CircleCI to build CMake from source as we don't have a newer image
  available. Also add some newly found dependencies (not sure how it was
  working without them before).

Test plan:

This change requires less testing than previous build related changes because
it deprecates most of the build types.

- The internal BUCK build is unaffected. No special testing.
- The semi-internal CMake build is gone. Use Nix.
- The Nix build for clang-15 and some tests are continuously tested in GitHub
  actions.
- Tested the set of Nix commands in the README. All work except the one that
  points to GitHub as this must be merged first.
- The existing CircleCI runs on Ubuntu 20.04 are maintained.
- Unable to test the new `test-report.yml` as it must be merged due to the
  permissions it needs. Will follow up with testing after this is merged. See:
  https://github.com/dorny/test-reporter?tab=readme-ov-file#recommended-setup-for-public-repositories

The list of exclusions for GitHub Actions/nix testing is currently very long, I
think 29% of the tests. This should be stable and reproducible though, and
likely needs deep changes to OI to fix. That's why fixes are excluded from this
PR. It's all to do with the forked drgn not being able to parse clang's newer
DWARF output, and can't be fixed by rolling back as we required a relatively
new libcxx.


.github/workflows/ci.yml
```diff
@@ -0,0 +1,54 @@
...
+          --parallel \
+          --no-compress-output \
+          --schedule-random \
+          --timeout 90 \
+          --repeat until-pass:3 \
+          --exclude-from-file ../../.github/workflows/tests_failing_under_nix.txt \
+          --output-junit results.xml
+    - name: upload results
+      uses: actions/upload-artifact@v4
+      if: success() || failure()
```
ajor, line 51: What other possibility is there?

CMakeLists.txt
```diff
@@ -33,75 +33,108 @@ find_package(gflags REQUIRED)
...
-
-# These glog executable targets still generate warnings - disable warnings for
-# them explicitly
-target_compile_options(demangle_unittest PRIVATE "-w")
-target_compile_options(logging_unittest PRIVATE "-w")
-target_compile_options(stl_logging_unittest PRIVATE "-w")
-target_compile_options(symbolize_unittest PRIVATE "-w")
-target_compile_options(utilities_unittest PRIVATE "-w")
+find_package(glog)
+if (NOT glog_FOUND)
```
ajor, line 45: As you yourself spotted yesterday, this isn't needed when using FIND_PACKAGE_ARGS

tools/config_gen.py
```diff
@@ -170,7 +171,10 @@ def pull_base_toml() -> typing.Dict:
 
     # Now, we need to replace any placeholders that might be present in the base toml file with the real verisons.
     user = getpass.getuser()
-    pwd = str(repo_path.resolve())
+    if "IN_NIX_SHELL" in os.environ and "src" in os.environ:
+        pwd = os.environ['src']
```
ajor, line 175: Who sets the `src` env var?

ajor (approved):

.github/workflows/ci.yml
JakeHillion: I think that's all of them. The implicit default is `if: success()` meaning the test analysis wouldn't get to run if any tests failed.

tools/config_gen.py
JakeHillion: Nix. I was surprised it isn't called `NIX_SRC_DIR` or something. I'm using this one so the placeholders get replaced with `/nix/store` store paths, rather than paths into your repo. This makes the output stable and also if the config file is ever a Nix output it'll have the correct dependencies.

CMakeLists.txt
JakeHillion: Yep! Other way round here though, it needed the check because those warning options (not entirely sure we need them but didn't want to make this any more complicated) aren't compatible with a `find_package` version of glog.

PR #504, opened 2024-08-15
build and test LLVM version 16

JakeHillion:
Add LLVM-16 to the Nix builds and CI matrix. Also add the option to explicitly
select an LLVM version so the CircleCI build can still work as before.

We should support at least LLVM-17 currently, but there appears to be a
significant performance regression in compiling the generated code for large
variants that I'm going to debug separately. The long term goal is to support
and test all versions 15 and up.

Test plan:
- CI

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -218,10 +218,15 @@ find_package(Boost REQUIRED COMPONENTS
 )
 message(STATUS "Linking Boost libraries: ${Boost_LIBRARIES}")
 
-### LLVM and Clang - Preferring Clang 15
-find_package(LLVM 15 REQUIRED CONFIG)
+### LLVM and Clang
+find_package(LLVM ${LLVM_REQUESTED_VERSION} REQUIRED CONFIG)
+
 message(STATUS "Found LLVM ${LLVM_PACKAGE_VERSION}")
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -218,11 +218,16 @@ find_package(Boost REQUIRED COMPONENTS
 )
 message(STATUS "Linking Boost libraries: ${Boost_LIBRARIES}")
 
-### LLVM and Clang - Preferring Clang 15
-find_package(LLVM 15 REQUIRED CONFIG)
+### LLVM and Clang
+find_package(LLVM ${LLVM_REQUESTED_VERSION} REQUIRED CONFIG)
```
ajor, line 222: Here's a nice little feature you could copy from bpftrace to make it easier for new users to set up. Make `LLVM_REQUESTED_VERSION` optional so they won't have to pass `-LLVM_REQUESTED_VERSION` if they only have one version installed:
```
if(${LLVM_REQUESTED_VERSION})
  find_package(LLVM ${LLVM_REQUESTED_VERSION} REQUIRED CONFIG)
else()
  find_package(LLVM REQUIRED CONFIG)
endif()
```

ajor (approved):

CMakeLists.txt
ajor: Oh, cool!

PR #507, opened 2024-08-22
add minimal support for aarch64

JakeHillion:
Summary:
OI is currently completely incompatible with anything except `x86_64`. Changing that generally is a big effort, but `oilgen` should work on other architectures. Begin adding some support for `aarch64` architecture.

This change sets up a file structure for architecture support. It pulls the 2 functions needed to make `Descs.{h,cpp}` architecture agnostic into architecture specific files for `x86_64` and `aarch64`. This enables `oilgen` (the binary) to build. At this stage `oilgen` is unable to generate working code for `aarch64`, but at least this is a step in the right direction.

Differential Revision: D61661524

### DIFF
--- a/oi/arch/Arch.h
+++ b/oi/arch/Arch.h
@@ -1,0 +1,29 @@
+/*
+ * Copyright (c) Meta Platforms, Inc. and affiliates.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *    http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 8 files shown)


oi/arch/Arch.h
```diff
@@ -0,0 +1,29 @@
...
+#include <cstdint>
+#include <optional>
+
+struct user_regs_struct;
+
+namespace oi::detail::arch {
+
+void setProgramCounter(user_regs_struct& regs, uintptr_t pc);
+
+std::optional<uintptr_t> getReturnValueAddress(const user_regs_struct&);
```
ajor, line 27: Why return an optional?

ajor (approved):

oi/arch/Arch.h
JakeHillion: I mainly did this to match the existing API, though it was never optional there either. This way it might be possible to part implement an architecture if it's ever difficult to get the return value/guess an arg position in future.

PR #518, opened 2024-09-18
Upgrade to Docusaurus v3

ttreyer:
## Summary
Dependabot raises regularly warnings about the dependencies used for the Docusaurus website. In particular, Docusaurus v2 is not maintained anymore and it uses NodeJS 16, which also reached its end of life.
This PR updates the website to use the current version of Docusaurus and removes multiple dev dependencies that were broken or unused.

## Test plan

The main breaking change comes from the MDX templates being updated to v3. Our MDX are simple enough to be compatible with the new version.
```
$ npx docusaurus-mdx-checker
[SUCCESS] All 8 MDX files compiled successfully!
```

Then doing a visual check through all our pages confirm the website stays identical to the previous version.
```
$ # Build and test the website locally
$ yarn build
[...]
$ yarn serve
```

### DIFF
--- a/website/package.json
+++ b/website/package.json
@@ -11,32 +11,20 @@
     "clear": "docusaurus clear",
     "serve": "docusaurus serve",
     "write-translations": "docusaurus write-translations",
-    "write-heading-ids": "docusaurus write-heading-ids",
-    "ci": "yarn lint && yarn format:diff",
-    "lint": "eslint --cache \"**/*.js\" && stylelint \"**/*.css\"",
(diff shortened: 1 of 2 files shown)


JakeHillion (approved):
Thanks for the update! I couldn't work out why this kept having dependency "issues" and this seems good. Can we hold off on merging this to avoid rebase issues with #517? I think it will go in slightly more easily that way.

GijsWeterings:
@JakeHillion actually let this one land in favor of #517 , per your internal request to D62962264 I've reverted the changes in that PR

PR #536, opened 2025-08-04
tests: add support for dynamically skipping tests based on llvm version

JakeHillion:
Some tests are useful but are either broken on individual LLVM versions or before/after an LLVM version. Add new options to the test TOML that let us skip tests dynamically based on too old/too new/explicitly skipped LLVM major versions.

Test plan:
- CI
- Used in stacked PRs


tyroguru (approved):
Excellent - thanks @JakeHillion!

